PR #144, opened 2026-04-27
Fix mobile autocapitalize handling

u473t8:
## Summary
- switch learner text fields from autocapitalize=off to autocapitalize=none
- set form-level autocapitalize/autocorrect defaults for add, search, edit, and lesson answer flows
- keep the previous minimal text-input UX scope otherwise unchanged

## Why
On real mobile Safari/WebKit behavior, autocapitalize=off may not suppress the initial shifted keyboard state reliably. Apple docs recommend none, and form-level attributes provide inherited defaults.

## Verification
- npx shadow-cljs compile node-test

Follow-up to #142

### DIFF
--- a/src/client/views/home.cljs
+++ b/src/client/views/home.cljs
@@ -47,6 +47,8 @@
          :autocapitalize "off"
          :autocomplete   "off"
          :autocorrect    "off"
+         :enterkeyhint   "next"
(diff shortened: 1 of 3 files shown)


u473t8:
Closing as superseded by #145, which has been merged.

PR #183, opened 2026-08-05
GH-179 Autocomplete: rank lemmas before aggregating translations

u473t8:
Closes #179.

## Defect

Short-prefix autocomplete cost ~100x more than long-prefix: native sqlite3 on the shipped `dictionary.sqlite`, prefix `fenster` 1 ms, prefix `f` 124 ms. The old `completions-sql` grouped and GROUP_CONCATed translations for every lemma in the prefix range (29,272 surface forms / 7,688 lemmas for `f`) before `ORDER BY rank` + `LIMIT 10` discarded all but ten — three TEMP B-TREEs sized by the range.

## Fix (issue option 1)

Select the winning ten lemmas first — distinct in-range lemma ids, pos filter, `ORDER BY rank DESC, value ASC LIMIT 10` — then compute `has_exact` as a point primary-key probe (`EXISTS` on `(normalized_form, lemma_id)`) and the translations `GROUP_CONCAT` for those ten only. Result columns, filter, ordering, and limit unchanged. Bind order in `completions` changed to (start, end, exact).

## Equivalence

Old and new SQL diffed byte-for-byte in sqlite3 for prefixes `f, fe, fen, fenster, sch, a, zu, ue`: **identical, all 8**.

Honest caveat on ordering: the sort key `(rank DESC, lemma ASC)` is not unique everywhere. Two result sets contain a tied pair — `zu` (lemma "zu" as adv and adj, same rank 999987) and `ue` (two "überfahren" verb lemmas, same rank 995118) — so the relative order of those two rows is formally unspecified in both the old and new query. Both engines happened to emit them in the same order here; no ties cross the LIMIT-10 boundary for any tested prefix, so the *set* of ten is deterministic.

## Benchmark (native sqlite3 3.44.4, warm cache, 5 runs, real time)

| Prefix | Old | New |
|---|---|---|
| `f` | 95–119 ms | 20–27 ms |
| `fe` | 14–16 ms | 3–4 ms |
| `fenster` | 0–1 ms | 0–1 ms |

~5x on the worst case. The residual ~22 ms for `f` is the unavoidable distinct-and-sort over 7.7k candidate lemmas (no rank index on the range); getting to ~1 ms would need issue option 2 (precomputed top lemmas per first letter), out of scope here.

## Verification

- `npx shadow-cljs compile app node-test`: both builds complete, **0 warnings**.
- `node target/node-tests.js`: 144 tests, 331 assertions, **0 failures**.
- Plainly: the node test suite does **not** exercise dictionary completions — no test touches `adapters.dictionary`. The equivalence proof above is the sqlite3 CLI diff, not app-level tests. In-app verification (WASM/OPFS path) pends the dev stand.

OpenSpec change `autocomplete-short-prefix-cost` validated and archived on this branch (spec `sqlite-dictionary-worker`).

### DIFF
--- a/src/client/adapters/dictionary.cljs
+++ b/src/client/adapters/dictionary.cljs
@@ -6,6 +6,18 @@
 
 
 (def ^:private completions-sql
+  "Top ten lemmas for a prefix range, cheap on short prefixes too.
+
+   Shape matters here. The inner SELECT DISTINCT collapses surface forms to
+   lemma ids before anything else: one lemma owns many in-range forms (Fenster,
+   Fensters, Fenstern...), and LIMIT counts rows, so without the collapse ten
+   rows would mean ten forms of maybe three lemmas. Rank then picks the winners
+   while the query still carries only (id, value, pos, rank) — no translations
+   joined, nothing concatenated. has_exact and translations are point lookups
+   for the surviving ten alone. The previous shape joined and GROUP_CONCATed
+   every in-range lemma — thousands on a one-letter prefix — and discarded all
+   but ten after sorting, which is why short prefixes cost ~100x more.
+   Measured in #179: prefix f 95-119 ms -> 20-27 ms."
   "WITH top AS (
      SELECT l.id, l.value, l.pos, l.rank
      FROM (SELECT DISTINCT lemma_id
--- a/src/client/adapters/dictionary.cljs
+++ b/src/client/adapters/dictionary.cljs
@@ -6,21 +6,28 @@
 
 
 (def ^:private completions-sql
-  "SELECT
-     l.id    AS lemma_id,
-     l.value AS lemma,
-     l.pos AS pos,
-     l.rank AS rank,
-     MAX(sf.normalized_form = ?) AS has_exact,
-     GROUP_CONCAT(DISTINCT t.value ORDER BY t.rank ASC) AS translations
-   FROM surface_forms sf
-   JOIN lemmas l ON l.id = sf.lemma_id
-   LEFT JOIN translations t ON t.lemma_id = l.id
-   WHERE sf.normalized_form >= ? AND sf.normalized_form <= ?
-     AND l.pos NOT IN ('conj', 'particle', 'pron', 'prep')
-   GROUP BY l.id
-   ORDER BY l.rank DESC, lemma ASC
-   LIMIT 10")
+  "WITH top AS (
+     SELECT l.id, l.value, l.pos, l.rank
+     FROM (SELECT DISTINCT lemma_id
+           FROM surface_forms
+           WHERE normalized_form >= ? AND normalized_form <= ?) m
+     JOIN lemmas l ON l.id = m.lemma_id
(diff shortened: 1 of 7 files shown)


src/client/adapters/dictionary.cljs
```diff
@@ -6,21 +6,28 @@
...
-   JOIN lemmas l ON l.id = sf.lemma_id
-   LEFT JOIN translations t ON t.lemma_id = l.id
-   WHERE sf.normalized_form >= ? AND sf.normalized_form <= ?
-     AND l.pos NOT IN ('conj', 'particle', 'pron', 'prep')
-   GROUP BY l.id
-   ORDER BY l.rank DESC, lemma ASC
-   LIMIT 10")
+  "WITH top AS (
+     SELECT l.id, l.value, l.pos, l.rank
+     FROM (SELECT DISTINCT lemma_id
```
u473t8, line 23: Why inner query instead of extending `where` condition?

src/client/adapters/dictionary.cljs
```diff
@@ -6,21 +6,28 @@
...
+     top.pos AS pos,
+     top.rank AS rank,
+     EXISTS (SELECT 1
+             FROM surface_forms sf
+             WHERE sf.normalized_form = ? AND sf.lemma_id = top.id) AS has_exact,
+     (SELECT GROUP_CONCAT(DISTINCT t.value ORDER BY t.rank ASC)
+      FROM translations t
+      WHERE t.lemma_id = top.id) AS translations
+   FROM top
+   ORDER BY top.rank DESC, lemma ASC")
```
u473t8, line 42: Can you explain, what makes the whole query efficient? Why it was constructed that way? Because for me a lot of things happen here.

src/client/adapters/dictionary.cljs
u473t8: Because one lemma owns many in-range surface forms (Fenster, Fensters, Fenstern → one lemma), and LIMIT counts rows: with a plain WHERE the ORDER BY … LIMIT 10 would keep ten *forms*, i.e. maybe three lemmas. The DISTINCT subquery collapses forms to lemma ids before ranking, so ten means ten lemmas. And "top ten by rank" cannot be said in a WHERE clause at all — a LIMIT needs its own query level to cut before the joins. Documented in the docstring now — 3b0dcd8.

src/client/adapters/dictionary.cljs
u473t8: The order of work is the whole trick: range scan on the PK → collapse forms to lemma ids → rank + LIMIT while a row is only (id, value, pos, rank) → and only for the surviving ten, two point lookups: has_exact as an EXISTS probe and translations as a scalar GROUP_CONCAT. The old shape joined translations onto *every* in-range lemma and concatenated thousands of groups (three TEMP B-TREEs in the plan), then threw away all but ten after sorting. Same answer now lives in the docstring — 3b0dcd8; numbers: prefix f 95–119 ms → 20–27 ms.

PR #184, opened 2026-08-05
GH-177 In-page measurement behind goog.DEBUG

u473t8:
Part of #177 — phase 1, the in-page half. Phase 2 (protocol-level traces) rides the Playwright suite, #169.

`window.__metrics()` in dev builds: renders, dispatches with nesting depth, frames, layout shift (input-caused entries filtered, CLS rule), long tasks, interactions over 100 ms. One guarded call in `main.cljs`; release builds eliminate the namespace.

Verified against the issue's own criterion — measurements must distinguish known cases. In headless Chrome (painting guaranteed): a saving dispatch renders exactly once; a deliberate 220 px layout shift scores 0.605; a 120 ms busy loop appears in long-tasks; 36 frames per 600 ms. Also measured and documented: an occluded window freezes frame- and paint-dependent metrics while dispatch and long-task counters keep working — relevant for anyone measuring against the visible Windows browser.

Tests: client 144 / 331, both builds 0 warnings.

### DIFF
--- a/src/client/instrumentation.cljs
+++ b/src/client/instrumentation.cljs
@@ -1,0 +1,109 @@
+(ns instrumentation
+  "In-page measurement for development. Answers, with numbers, what the
+   browser did: how many times the app rendered, how many dispatches ran and
+   how they nested, whether layout shifted, which tasks ran long, which
+   interactions were slow.
+
+   Installed only under goog.DEBUG — a release build eliminates the whole
+   namespace. Read from the page as `window.__metrics()`, reset with
+   `window.__metricsReset()`.
+
+   Honest limits, measured before this existed: the frame counter stops when
+   the window is occluded even though visibilityState stays \"visible\", and a
+   render here means one store notification — the browser still paints at most
+   once per frame regardless of how many happened inside it."
+  (:require
+   [nexus.registry :as nxr]))
+
+
+(defonce ^:private zero-metrics
+  {:dispatches   0
+   :frames       0
+   :layout-shift 0.0
+   :long-tasks   []
+   :nested-dispatches 0
+   :renders      0
+   :slow-interactions []})
+
+
+(defonce ^:private metrics (atom zero-metrics))
+
+
+(defonce ^:private dispatch-depth (volatile! 0))
+
+
+(defn- observe!
+  "Watches one PerformanceObserver entry type, quietly skipping the ones this
+   browser does not support."
+  [entry-type f options]
+  (try
+    (let [observer (js/PerformanceObserver.
+                    (fn [entries _]
+                      (doseq [entry (.getEntries ^js entries)]
+                        (f entry))))]
+      (.observe observer (clj->js (merge {:type entry-type :buffered true} options)))
+      observer)
+    (catch :default _ nil)))
+
+
+(defn- count-frames!
+  []
+  (js/requestAnimationFrame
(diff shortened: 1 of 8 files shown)


src/client/instrumentation.cljs
```diff
@@ -0,0 +1,109 @@
...
+
+(defonce ^:private dispatch-depth (volatile! 0))
+
+
+(defn- observe!
+  "Watches one PerformanceObserver entry type, quietly skipping the ones this
+   browser does not support."
+  [entry-type f options]
+  (try
+    (let [observer (js/PerformanceObserver.
```
u473t8: I am not familiar with this Web API, what does constructor callback mean?

src/client/instrumentation.cljs
```diff
@@ -0,0 +1,109 @@
...
+
+  ;; Shifts the user did not cause: hadRecentInput filters out reflows that
+  ;; follow the user's own click or keystroke, the same rule CLS scoring uses.
+  (observe! "layout-shift"
+            (fn [entry]
+              (when-not (.-hadRecentInput ^js entry)
+                (swap! metrics update :layout-shift + (.-value ^js entry))))
+            {})
+
+  (observe! "longtask"
```
u473t8, line 89: Explain please, why the observer object, create by the observe! functions is outlives the function scope? I also see that the function returns observer object, but it is not used anywhere

src/client/instrumentation.cljs
```diff
@@ -0,0 +1,109 @@
...
+            (fn [entry]
+              (swap! metrics update
+                :slow-interactions
+                conj
+                {:duration (.-duration ^js entry)
+                 :type     (.-name ^js entry)}))
+            {:durationThreshold 100})
+
+  (set! (.-__metrics js/window) (fn [] (clj->js @metrics)))
+  (set! (.-__metricsReset js/window)
```
u473t8, line 113: Can you explain how these fields are used? Why we assign a function to `__metrics`?

src/client/instrumentation.cljs
u473t8: PerformanceObserver is a subscribe API: the constructor takes a callback, and the *browser* invokes it with a batch of entries whenever it records new ones (a long task finished, a layout shifted). `.observe` only declares which entry type to watch and whether to replay already-buffered entries. So the callback is not called by us at all — it is the delivery channel. Now said in the docstring — ae40471.

src/client/instrumentation.cljs
u473t8: Fair catch — the return was decoration. The object survives not because anyone holds it: the platform keeps a *registered* observer alive until `.disconnect()`, it sits in the browser's observer list. And since these probes watch for the whole page lifetime, nothing ever tears them down, so holding a reference had no use. The return is removed and the lifetime rule is in the docstring — ae40471.

src/client/instrumentation.cljs
u473t8: They are the page-facing API: CDP evals and future Playwright specs call `window.__metrics()` to read a snapshot and `__metricsReset()` between scenarios — that is how the #184 verification numbers were taken. A function rather than a value because each call converts the atom's *current* state (`(clj->js @metrics)`); a value assigned once would be frozen at install time. Comment added — ae40471.

PR #196, opened 2026-08-05
GH-192 Configurable database path via LEARNING_APP_DB_PATH

u473t8:
Depends-on: #201

Closes #192.

## What

`db-spec` in `src/backend/core.clj` now reads `LEARNING_APP_DB_PATH`, defaulting to `app.db` — same pattern as `port` / `LEARNING_APP_PORT`.

- `db-spec` remains the single home of the app database spec: migrations and handlers take it as an argument; dictionary tools and tests build their own specs for their own files (verified by grep).
- Production unchanged: the service runs with `WorkingDirectory=/opt/learning-app` and no `LEARNING_APP_DB_PATH`, so the default resolves to `/opt/learning-app/app.db` exactly as before. Service file untouched.
- OpenSpec: new `backend-configuration` capability, change archived as `2026-08-05-configurable-db-path` in the same commit.

## Verification

Backend tests:

```
Ran 11 tests containing 26 assertions.
0 failures, 0 errors.
```

Env var proof — migrations booted against the pointed path:

```
$ rm -f /tmp/probe-192.db && LEARNING_APP_DB_PATH=/tmp/probe-192.db clj -M:dev -e "(require 'core 'migrations) ..."
db-spec: {:dbname /tmp/probe-192.db, :dbtype sqlite}
exists: true
-rw-r--r-- 1 u473t8 u473t8 77824 Aug  5 17:49 /tmp/probe-192.db
```

Default without the var:

```
default db-spec: {:dbname app.db, :dbtype sqlite}
```

Not verified: running production service with the new jar (backend-only worktree, dev stand untouched).

### DIFF
--- a/src/backend/core.clj
+++ b/src/backend/core.clj
@@ -68,7 +68,8 @@
 
 
 (def db-spec
-  {:dbtype "sqlite" :dbname "app.db"})
+  {:dbname (or (System/getenv "LEARNING_APP_DB_PATH") "app.db")
+   :dbtype "sqlite"})
 
 
 (defmacro on-connection
--- a/openspec/changes/archive/2026-08-05-configurable-db-path/.openspec.yaml
+++ b/openspec/changes/archive/2026-08-05-configurable-db-path/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven-with-adr
+created: 2026-08-05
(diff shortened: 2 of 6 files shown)


src/backend/core.clj
```diff
@@ -68,7 +68,8 @@
 
 
 (def db-spec
-  {:dbtype "sqlite" :dbname "app.db"})
+  {:dbname (or (System/getenv "LEARNING_APP_DB_PATH") "app.db")
```
u473t8, line 71: convention for app envs is `LEARNING_APP__XXX`, e.g. `LEARNING_APP__DB_PATH`

PR #197, opened 2026-08-05
GH-182 Refuse account ids whose userdb already exists

u473t8:
Closes #182.

Account ids are `users` row ids and userdb names derive from them, but CouchDB outlives `app.db` resets — a recycled id silently inherited the previous owner's data (112 foreign documents on a fresh account, seen live). The existence check now shares a transaction with the row insert: refusal leaves the id unspent and the stale userdb intact as evidence; the route answers 503 and logs `::recycled-id-refused`. Deleting the stale userdb stays the operator's call.

Verification, all live on the stand: stale `userdb-2` → HTTP 503, `users` count unchanged, event in the log; after operator deletion → HTTP 200, id 2, `userdb-2` created with `{"roles":["u:2"]}`. Tests 13/32 green, including rollback of the refused row. Also fixed en route: the id extraction now pins its own result-set builder instead of inheriting whichever options the caller's connection carries.

Known cost, stated: the invite burns before the refusal is discovered — after clearing the conflict the operator mints a new one. Acceptable for a corruption-detection path.

### DIFF
--- a/openspec/changes/archive/2026-08-05-recycled-id-guard/proposal.md
+++ b/openspec/changes/archive/2026-08-05-recycled-id-guard/proposal.md
@@ -1,0 +1,13 @@
+# Recycled account ids do not inherit userdbs
+
+## Why
+
+Account ids are `users` row ids and the CouchDB database name derives from them, but the two stores have independent lifetimes: an `app.db` reset restarts id allocation while `userdb-*` survive. Seen live (#182): a freshly provisioned account received id 1 and its first sync pass pulled 112 documents belonging to the previous owner of `userdb-1`.
+
+## What changes
+
+`create-account!` refuses an id whose userdb already exists — the row insert and the existence check share a transaction, so a refusal leaves the id unspent and the stale userdb intact as evidence. The provision route answers 503 and logs `::recycled-id-refused`. Deleting the stale userdb is the operator's call, not the code's.
+
+## Impact
+
+`create-account!`, the provision route, two tests, live verification on the dev stand. Known cost: the invite is burned before the refusal is discovered — the operator mints a new one after clearing the conflict.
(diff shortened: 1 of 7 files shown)


openspec/changes/archive/2026-08-05-recycled-id-guard/proposal.md
```diff
@@ -0,0 +1,13 @@
+# Recycled account ids do not inherit userdbs
+
+## Why
+
+Account ids are `users` row ids and the CouchDB database name derives from them, but the two stores have independent lifetimes: an `app.db` reset restarts id allocation while `userdb-*` survive. Seen live (#182): a freshly provisioned account received id 1 and its first sync pass pulled 112 documents belonging to the previous owner of `userdb-1`.
+
+## What changes
+
+`create-account!` refuses an id whose userdb already exists — the row insert and the existence check share a transaction, so a refusal leaves the id unspent and the stale userdb intact as evidence. The provision route answers 503 and logs `::recycled-id-refused`. Deleting the stale userdb is the operator's call, not the code's.
```
u473t8, line 9: Can we just check intergrity of couch db and sqlite? I mean check that all user bases in couch db have corresponding rows in sqlite? And, in case if no, tombstone these user dbs?

u473t8: Yes to the check, no to the auto-tombstone — filed as #205. The dangerous false positive: SQLite restored from an older backup. Every account created after that backup has a live, legitimate userdb and no row — auto-tombstoning at that moment destroys real user data exactly when the system is most fragile (and a half-finished #201 migration looks the same). So: a reconciliation *report* naming the orphans, disposal stays a human decision. The per-provision guard here keeps its separate job — it is race-free at the moment of harm, inside the same transaction, where a periodic sweep always has a window.

PR #198, opened 2026-08-05
GH-189 Verified atomic jar swap on deploy

u473t8:
Closes #189.

## How the jar lands in /opt/learning-app today (diagnosis)

`deploy-app.yml` (manual `workflow_dispatch`) builds the uberjar, copies it to `target/learning-app.jar.tmp`, and `appleboy/scp-action` (tar-over-ssh) extracts it as `/opt/learning-app/learning-app.jar.tmp`. A follow-up `appleboy/ssh-action` step runs `test -f` then `mv -f` over `/opt/learning-app/learning-app.jar`. On the server, `learning-app-restart.path` (`PathChanged=` on the live jar) fires `learning-app-restart.service`, which restarts `learning-app-run.service`. There is no in-place copy of the live jar: the write-aside + rename core has been in place since GH-64, and `mv` within `/opt/learning-app` is rename(2) on one filesystem.

Two torn-jar paths remained:

1. **Fixed shared temp name.** Two concurrent dispatch runs both write `/opt/learning-app/learning-app.jar.tmp`. Run A's `mv -f` can grab the file while run B is still uploading it — the rename is atomic, but it atomically installs a torn jar, `PathChanged` fires, and the service restarts on it. `Restart=always` then crash-loops.
2. **No integrity gate.** `test -f` passes on a truncated upload; nothing proved the tar-over-ssh transfer delivered the artifact whole before it got swapped in.

## Change

- Upload path is per-run unique: `learning-app.jar.${{ github.run_id }}.tmp`, still inside `/opt/learning-app` so the rename never crosses filesystems.
- Build records the jar's sha256; the server-side step verifies it against the uploaded temp file (`sha256sum -c`) and aborts before the swap on mismatch or missing file (subsumes `test -f`).
- After a successful swap, leftover `learning-app.jar.*.tmp` older than an hour are deleted, so aborted runs do not accumulate uberjars.

Restart mechanism and everything else in the deploy untouched.

## Verification (honest scope)

- Workflow YAML parses (`python3 -c "import yaml; yaml.safe_load(...)"`).
- The server-side script was dry-run verbatim against a simulated `/opt/learning-app`: happy path verifies then renames then cleans a 2h-old stale temp; a torn upload fails `sha256sum -c` **before** the `mv` with the live jar untouched; a missing upload fails the step.
- **Not proven here: the real transfer path.** The proof is the next real deploy. Watch for: the ssh step printing `…learning-app.jar.<run_id>.tmp: OK` from `sha256sum -c`, exactly one service restart after the `mv`, app healthy on the new jar, and no `learning-app.jar.*.tmp` left in `/opt/learning-app` afterwards.

OpenSpec: capability `deploy-pipeline` added, change `2026-08-05-atomic-artifact-swap` archived in this commit (task 5 — watching the next real deploy — deliberately left open).

### DIFF
--- a/.github/workflows/deploy-app.yml
+++ b/.github/workflows/deploy-app.yml
@@ -38,7 +38,10 @@ jobs:
       - run: rm -rf resources/public/js/app
       - run: npx shadow-cljs release app
       - run: clojure -T:build uber
-      - run: cp target/learning-app.jar target/learning-app.jar.tmp
+      # Unique per-run temp name: concurrent deploys never share an upload path.
+      - run: cp target/learning-app.jar 'target/learning-app.jar.${{ github.run_id }}.tmp'
+      - id: jar
+        run: echo "sha256=$(sha256sum target/learning-app.jar | cut -d' ' -f1)" >> "$GITHUB_OUTPUT"
 
       - uses: actions/cache/save@v6.1.0
         if: ${{ always() && steps.cache-clojure.outputs.cache-hit != 'true' }}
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy-app.yml
```diff
@@ -38,7 +38,10 @@ jobs:
       - run: rm -rf resources/public/js/app
       - run: npx shadow-cljs release app
       - run: clojure -T:build uber
-      - run: cp target/learning-app.jar target/learning-app.jar.tmp
+      # Unique per-run temp name: concurrent deploys never share an upload path.
+      - run: cp target/learning-app.jar 'target/learning-app.jar.${{ github.run_id }}.tmp'
+      - id: jar
+        run: echo "sha256=$(sha256sum target/learning-app.jar | cut -d' ' -f1)" >> "$GITHUB_OUTPUT"
```
u473t8, line 47: Just curious, can you add a github comment with the link to the "$GITHUB_OUTPUT" documentation?

u473t8 (commented):
LGTM. How do you think, by the way, is it already time to think about blue-green deployment?

.github/workflows/deploy-app.yml
u473t8: Added right above the line — with the contract in one sentence and the link: cfba080.

PR #199, opened 2026-08-05
GH-188 Graceful shutdown of the backend on SIGTERM

u473t8:
Closes #188.

## Defect

systemd stop sends SIGTERM; the JVM died mid-request — listening socket vanished without draining, in-flight responses were cut, request-held SQLite connections severed.

## Mechanism

- `defonce`-guarded JVM shutdown hook calls `stop-server!` and logs "Server stopped". One hook per JVM: the dev reload path cannot register a duplicate, and the hook only runs at JVM shutdown, so it cannot fight `restart-server!`.
- `stop-server!` passes a 5 s drain timeout to http-kit `server-stop!`: socket closes first, in-flight requests finish. Draining is all the SQLite cleanup there is — connections are per-request (`with-open` in `on-connection`), stated in the hook comment.
- `restart-server!` delegates to `stop-server!` so both paths share the drain semantics.
- OpenSpec: new `backend-operations` capability, change archived as `2026-08-05-graceful-backend-shutdown`.

## Proven

- New test `stopping-the-server-drains-in-flight-requests`: request in flight when `stop-server!` runs completes with 200, socket refuses afterwards, handle released. Suite: 12 tests / 30 assertions, 0 failures (was 11/26 before this change).
- Live SIGTERM on free port 8199 (temp db in worktree, dev stand untouched): "Server stopped" logged, process exited, port closed, no stack traces.

## Not proven

- Behavior under systemd itself (unit `TimeoutStopSec` interplay, real production stop) — only a live SIGTERM on the stand can show that; the local proof used the same signal path but not systemd.

### DIFF
--- a/src/backend/core.clj
+++ b/src/backend/core.clj
@@ -496,24 +496,44 @@
   (reset! server (server/run-server app {:port port :legacy-return-value? false})))
 
 
-#_{:clojure-lsp/ignore [:clojure-lsp/unused-public-var]}
+(def ^:private drain-timeout-ms
+  "How long a stopping server keeps serving in-flight requests after the
+   listening socket has closed."
+  5000)
 
 
 (defn stop-server!
   []
-  (when (some? @server)
-    (when-let [stopping-promise (server/server-stop! @server)]
-      @stopping-promise
+  (when-some [running @server]
+    (when-some [stopping (server/server-stop! running {:timeout drain-timeout-ms})]
+      @stopping
       (reset! server nil))))
+
+
+#_{:clj-kondo/ignore [:unused-private-var]}
+
+
+(defonce ^:private shutdown-hook
+  ;; systemd stop is SIGTERM; without this hook the JVM dies mid-request.
+  ;; Stopping the server closes the listening socket first, then drains
+  ;; in-flight requests — and draining is all the cleanup there is: SQLite
+  ;; connections are per-request (`with-open` in `on-connection`), so the last
+  ;; response closes the last one. `defonce` keeps the dev reload path from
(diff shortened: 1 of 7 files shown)


src/backend/core.clj
```diff
@@ -496,24 +496,44 @@
...
+(defonce ^:private shutdown-hook
+  ;; systemd stop is SIGTERM; without this hook the JVM dies mid-request.
+  ;; Stopping the server closes the listening socket first, then drains
+  ;; in-flight requests — and draining is all the cleanup there is: SQLite
+  ;; connections are per-request (`with-open` in `on-connection`), so the last
+  ;; response closes the last one. `defonce` keeps the dev reload path from
+  ;; registering a second hook; the hook itself only ever runs at JVM
+  ;; shutdown, so it cannot fight `restart-server!`.
+  (let [hook (Thread. ^Runnable
+                      (fn []
```
u473t8, line 544: Just curious, how clojure convert lambda to Runnable instance? Does it work because of the type hint?

u473t8 (commented):
LGTM

src/backend/core.clj
u473t8: No conversion happens — a Clojure fn already *is* a Runnable: every fn extends AFn, which implements both java.lang.Runnable and java.util.concurrent.Callable (run/call just invoke the fn). So (Thread. (fn ...)) works with no hint at all. The ^Runnable hint does something else: Thread has overloaded constructors, and the hint tells the *compiler* which one to bind at compile time instead of resolving by reflection at runtime. Behavior identical, one reflective lookup saved — and the reflection warning silenced, which this codebase treats as a build requirement (*warn-on-reflection* is on).

PR #200, opened 2026-08-05
GH-190 Harden systemd services

u473t8:
Closes #190.

## What

All four `[Service]` units under `infra/production/etc/systemd/system/` get a sandbox block sized to what each service was observed (in unit files, `admin-setup.sh`, `backup.sh`, `import.sh`, and backend source) to actually need. Timers and the `.path` unit have no `[Service]` section — nothing to harden there.

## Applied

| Directive | run | dictionary-import | backup-db | certbot | restart |
|---|---|---|---|---|---|
| NoNewPrivileges | ✓ (was) | ✓ (was) | ✓ (was) | ✓ | ✓ |
| PrivateTmp | ✓ (was) | ✓ (was) | ✓ (was) | ✓ | ✓ |
| ProtectHome | ✓ (was) | ✓ (was) | ✓ | ✓ | ✓ |
| ProtectSystem | strict | strict | strict | full | strict |
| ReadWritePaths | /opt/learning-app (SQLite app.db in cwd) | /var/lib/learning-app/dictionary | /var/backups/learning-app, /var/lib/dbmaintainer (borg home) | /etc/letsencrypt, /etc/nginx | — |
| PrivateDevices | ✓ | ✓ | ✓ | — | ✓ |
| ProtectKernelTunables/Modules/Logs, ProtectControlGroups, ProtectClock, ProtectHostname | ✓ | ✓ | ✓ | ✓ | ✓ |
| RestrictSUIDSGID, RestrictNamespaces, RestrictRealtime, LockPersonality | ✓ | ✓ | ✓ | ✓ | ✓ |
| RestrictAddressFamilies | INET INET6 UNIX | INET INET6 UNIX | INET INET6 UNIX (ssh to borg repo) | INET INET6 UNIX (ACME + D-Bus) | UNIX only |
| CapabilityBoundingSet= (empty) | ✓ | ✓ | ✓ | — | ✓ |
| SystemCallArchitectures=native, SystemCallFilter=@system-service | ✓ | ✓ | ✓ | ✓ | ✓ |
| MemoryDenyWriteExecute | off | off | off | off | **on** |

StateDirectory/LogsDirectory stay writable under `ProtectSystem=strict` automatically; that covers `/var/lib/learning-app` (including the `-wal`/`-shm` write access sqlite3 needs even for a read-only `.backup` of a WAL database).

## Not applied, reason

- **MemoryDenyWriteExecute** on run/dictionary-import: HotSpot JIT maps writable+executable pages — enabling it kills the JVM. On backup-db/certbot: Python with cffi; libffi closure trampolines need W+X. Only `learning-app-restart` (plain `systemctl`) enforces it.
- **CapabilityBoundingSet** narrowing on certbot: the renewer and deploy hook chown/chmod cert files as root; the minimal capability set is unverified against a real renewal, so caps stay broad rather than guessed.
- **ProtectSystem=strict** on certbot: certbot's full write surface under `/var` and `/run` is not enumerable from code with confidence; `full` + explicit `/etc` holes is the honest bound.
- **ProtectProc/ProcSubset**: JVM container/cgroup detection reads non-pid `/proc` entries; survival unverified, left out.
- **PrivateUsers, PrivateNetwork, IPAddress{Allow,Deny}, UMask tightening**: either break declared needs (network, group-writable drop dirs) or are unverifiable here; left out.

## Verification

**Done here (WSL, systemd 255):** `systemd-analyze verify` over all eight units — passes; the only diagnostic is `/usr/bin/certbot is not executable`, an artifact of certbot not being installed in the dev environment. Verify also caught a real bug in the first draft: systemd ignores inline comments on directive lines silently — every directive would have been dropped. All comments are full lines now. The infra-checks deb build proves packaging only, not unit semantics.

**Not possible here, staged for the server (in this order):**
1. `systemd-analyze security learning-app-run` — record the exposure score **before** deploying, then after; expect a drop from ~9.6 ("UNSAFE") territory to ~4 or below.
2. Deploy, `systemctl daemon-reload`, `systemctl restart learning-app-run`, `systemctl status` clean, journal free of seccomp/EPERM kills.
3. Full request path: `curl -f https://sprecha.de/` and `curl -f https://sprecha.de/auth/check` (exercises SQLite write, CouchDB proxy-auth signing, credential reads).
4. Migration run: restart performs startup migrations against `app.db` — confirm in journal that they complete under `ProtectSystem=strict`.
5. `systemctl start learning-app-backup-db` once, check borg reaches the repo over ssh and `/var/backups/learning-app/db.sqlite` passes `.selftest`.
6. `certbot renew --dry-run` inside the unit (`systemd-run --unit=... -p ...` or a timer dry-run) before the real 03:30 window.

If any service trips seccomp or a read-only path, the journal names the syscall/path; loosen that single directive rather than reverting the block.

### DIFF
--- a/infra/production/etc/systemd/system/learning-app-backup-db.service
+++ b/infra/production/etc/systemd/system/learning-app-backup-db.service
@@ -17,5 +17,41 @@ Environment=BORG_PASSPHRASE_PATH=%d/borg-passphrase
 Environment=LEARNING_APP_DB_BACKUP_PATH=/var/backups/learning-app/db.sqlite
 
 # Security settings
+#
+# What this service actually needs: sqlite3 .backup of the app database into
+# /var/backups/learning-app, then borg over ssh to the remote repository.
+# Borg keeps its cache/config in the dbmaintainer home (/var/lib/dbmaintainer,
+# not under /home, so ProtectHome does not block it).
+#
+# No privilege gain on execve; nothing here elevates.
 NoNewPrivileges=yes
 PrivateTmp=yes
+ProtectHome=yes
+# Filesystem read-only except ReadWritePaths below. StateDirectory keeps
+# /var/lib/learning-app writable under strict — sqlite3 needs write access to
+# the source -wal/-shm files even when only reading a WAL database.
+ProtectSystem=strict
+# Local backup target plus borg cache/config/ssh state in the home directory.
+ReadWritePaths=/var/backups/learning-app /var/lib/dbmaintainer
+PrivateDevices=yes
+# Kernel interfaces read-only or unreachable.
+ProtectKernelTunables=yes
+ProtectKernelModules=yes
+ProtectKernelLogs=yes
+ProtectControlGroups=yes
+ProtectClock=yes
+ProtectHostname=yes
+# No setuid/setgid files, namespaces, realtime, or personality changes.
+RestrictSUIDSGID=yes
+RestrictNamespaces=yes
+RestrictRealtime=yes
+LockPersonality=yes
+# ssh to the borg repository; AF_UNIX for journald and NSS lookups.
+RestrictAddressFamilies=AF_INET AF_INET6 AF_UNIX
+# Unprivileged user: zero capabilities needed.
+CapabilityBoundingSet=
+# Native-ABI syscalls only, filtered to the standard service set.
+SystemCallArchitectures=native
+SystemCallFilter=@system-service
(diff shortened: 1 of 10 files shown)


u473t8 (commented):
I guess it looks good? The matter is, I already forgot what are we trying to achieve here. I remember I was going to read documentation and learn the purpose of all these options. But now it is incomprehensible for me, and I can't judge :)

u473t8 (commented):
Let's use it as a study case to learn more about systemd

PR #203, opened 2026-08-05
GH-202 Depends-on lines become a merge gate

u473t8:
Closes #202.

`Depends-on: #N` in a PR body now gates the merge: the check fails while any referenced node is open — or does not exist, so a typo cannot unlock anything. A companion job re-runs open dependents after each push to master; the one blind spot (an issue closed by hand, no push) is documented in the workflow and covered by `workflow_dispatch`.

The PR body enters the script through env, never template interpolation — safe on a public repo, fork PRs run with a read-only token.

Also here: `integration.yml` paths widen to `.github/workflows/**` — the old filter listed only itself, so this very PR would have hung on the required-but-never-reporting Client Tests. The broader form of that trap (docs-only PRs vs required path-filtered checks) is filed separately.

Verified locally against live issues: no refs → green; #201 (open) → red; #168 (closed) → ok; #99999 (missing) → red.

## Rollout after merge, in order

1. Add context "Depends-on" to required status checks.
2. Probe: fire the check on an open PR, watch the button die.
3. Retrofit `Depends-on: #201` onto #196 and drop its draft — the wall holds by itself.

### DIFF
--- a/.github/workflows/depends-on.yml
+++ b/.github/workflows/depends-on.yml
@@ -1,0 +1,80 @@
+# Turns `Depends-on: #N` lines in a PR body into a merge gate: the check fails
+# while any referenced issue or PR is open, and branch protection does the
+# enforcing. The companion job re-runs the check on open dependents after a
+# push to master, so they go green without a manual poke. One blind spot: an
+# issue closed *without* a push (by hand) leaves dependents red until the next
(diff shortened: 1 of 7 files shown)


u473t8 (commented):
LGTM, I am a noob regarding bash scripting and understanding what data is available in a run.

PR #208, opened 2026-08-06
GH-205 orphan-userdb reconciliation report

u473t8:
Closes #205.

Reconciliation report of CouchDB `userdb-N` databases without a matching `users` row in SQLite. Report only — auto-tombstoning is forbidden (restore-from-older-backup makes orphans live, legitimate data); disposal is a human decision. Rows without userdbs are a different signal and are not reported.

- `db/all-dbs` in lib/db (GET /_all_dbs, admin conn)
- `reconciliation/orphan-userdbs` (the diff) and `reconciliation/report!` (logs via telemere, returns the list — REPL operator entry point, like `mint-grant!`)
- Startup hook in `-main` after migrations; CouchDB down → warning, boot continues
- Tests: orphan named exactly; matched stores → empty; rows without dbs not reported; CouchDB down → nil, no throw. Suite: 18 tests / 40 assertions green (baseline 14/36)

**Pending live verification** on the dev stand (not done from this worktree — shared stand untouched): stage an orphan userdb → startup log and REPL report name it and nothing else. OpenSpec task 5 tracks it.

### DIFF
--- a/lib/db/src/db.cljc
+++ b/lib/db/src/db.cljc
@@ -183,6 +183,20 @@
 
 
 #?(:clj
+   (defn all-dbs
+     "Returns the names of all databases on the server.
+
+      https://docs.couchdb.org/en/stable/api/server/common.html#all-dbs"
+     ([]
+      (all-dbs conn))
+     ([conn]
+      (let [response (request-sync conn {:method :get :url "_all_dbs"})]
+        (if (= (:status response) 200)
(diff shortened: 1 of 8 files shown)


u473t8 (commented):
That's fun, but why do not we use user token as user id? As far as I remember we already use it sqlite db, and it uniquely identifies the user

PR #209, opened 2026-08-06
GH-201 The database adopts its configured home at startup

u473t8:
Closes #201. Closes #192.

Production's EnvironmentFile has pointed at `/var/lib/learning-app/db.sqlite` all along while the app wrote `/opt/learning-app/app.db` — so backups archive a file nobody writes, and merging the env-reading change alone (#196) would strand every account on an empty database at the next deploy.

The design inverts #201's original direction: no move step in the deploy pipeline. The backend adopts the legacy file itself at startup, in the same artifact that reads `LEARNING_APP_DB_PATH` — the two halves cannot ship apart, which is what closes the outage window. Mechanics: copy under a temp name → atomic rename → delete the legacy trio; crash mid-copy retries next boot with the legacy still authoritative; an existing target is never clobbered (rollback must not cost newer data) — both-exist logs `legacy-database-left-behind` for the operator.

Infra turned out to need zero changes: `StateDirectory=learning-app` already creates and unlocks the directory under the #200 sandbox, environment.d already carries the path, `backup.sh` already points at the target. First commit is #196's env-reading change, cherry-picked with authorship preserved; PR #196 is superseded and will be closed.

Verified: 18 tests / 45 assertions (move with wal, no-clobber, no-legacy, dev same-path); end-to-end simulation from a scratch cwd — marker row survives, boot 200, `database-adopted` in the log, graceful stop. Pending the real deploy: adoption event in the journal, `/auth/check` 200 with an existing token, a backup containing real rows (that last one also settles half of #206).

Rollback caveat documented in the runbook: a pre-#201 jar redeployed after adoption starts empty in `/opt` — 401s until rolling forward; data stays safe at the new home.

### DIFF
--- a/docs/ops/runbook.md
+++ b/docs/ops/runbook.md
@@ -74,6 +74,18 @@ curl -sf http://127.0.0.1:5984/ >/dev/null
 curl -sf --resolve sprecha.de:443:127.0.0.1 https://sprecha.de/db/dictionary-db/dictionary-meta >/dev/null
 ```
 
+## Server database home
+
+The server SQLite lives at `/var/lib/learning-app/db.sqlite` (`LEARNING_APP_DB_PATH`
+from environment.d; the directory is systemd's `StateDirectory`). Builds carrying
+#201 adopt a legacy `/opt/learning-app/app.db` on first boot: move it, log
+`database-adopted`, and refuse to clobber an existing target (`legacy-database-left-behind`
+in the log means both files exist — resolve by hand, the running app is on the target).
+
+Rollback caveat: redeploying a pre-#201 jar after adoption starts an empty
+database in `/opt` — tokens then 401 until a #201+ jar is redeployed. The data
+itself stays safe at the new home.
+
 ## Manual admin steps (bootstrap + secrets + one-off ops)
 
 Run the admin setup script (idempotent, safe to re-run).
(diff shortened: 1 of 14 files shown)


docs/ops/runbook.md
```diff
@@ -74,6 +74,18 @@ curl -sf http://127.0.0.1:5984/ >/dev/null
 curl -sf --resolve sprecha.de:443:127.0.0.1 https://sprecha.de/db/dictionary-db/dictionary-meta >/dev/null
 ```
 
+## Server database home
+
+The server SQLite lives at `/var/lib/learning-app/db.sqlite` (`LEARNING_APP_DB_PATH`
```
u473t8: Forgot to mention, all app envs should match the pattern `LEARNING_APP__XXX`, for example `LEARNING_APP__DB_PATH`

u473t8: Renamed — 7779702: code, environment.d, both docs, the change record; the env-boot check answers with the pointed path under the new name. Deploy-order safe both ways: whichever artifact lands first, the missing name degrades to today's app.db behavior, so no coordination needed. Two deliberate leftovers: backup.sh keeps the old name (its := default covers the same path, and the file belongs to #211's open PR — no cross-PR conflict), and the remaining four single-underscore envs (PORT, DB_AUTH_SECRET, DB_AUTH_SECRET_PATH, DB_BACKUP_PATH) are filed as the full-convention sweep — one of them needs care: the secret's name is wired through the systemd unit, and unit and jar ship in different artifacts, so a naive rename opens a boot-refusal window.

u473t8 (commented):
Did you make sure the correct env name is used across infrastructure configs?

PR #210, opened 2026-08-06
GH-169 Headless Playwright suite replacing ad-hoc CDP checks

u473t8:
Closes #169.

## What

Playwright suite in `test/browser/`, headless inside WSL against the system Google Chrome (`channel: "chrome"` — verified to launch, no browser download) and a locally started backend (`LEARNING_APP_PORT=8301`), per the recorded decision: no mirrored networking, the CDP skill keeps the visible Windows Chrome. Config: baseURL `http://localhost:8301`, `trace: retain-on-failure`. Role/label locators and auto-waiting assertions only — no `waitForTimeout`, no CSS-class locators. `npm run test:browser` assumes the backend is already up (documented in `test/browser/README.md`).

## Honest scope notes

- **Scenario 1** (word add pushes to server) reduces to *the word persists*: visible in the words list and surviving a full reload. No CouchDB runs here, so no replication request exists to assert on.
- **Scenario 3** exercises the **install-guide dialog**, not the sync menu. The sync button (`aria-label "Синхронизация"`) renders only when `:app/account-id` exists, and provisioning requires burning an invite grant (`/api/identity/provision`, ADR-0006) — absent in a fresh environment. The spec asserts the sync button's absence to document this, then opens the install guide under an iOS user agent (the only path where the install click shows the guide rather than the native prompt), verifies an in-card click does *not* dismiss, and a backdrop click does.
- **Out of scope:** CI wiring (deferred by the issue), service worker/offline, sync/CouchDB flows.

OpenSpec: new capability `browser-test-suite`, change archived as `2026-08-06-headless-playwright-suite` in the same commit.

## Verification

```
$ npm run test:browser
Running 3 tests using 3 workers

  ✓  2 test/browser/dialog-backdrop.spec.js:14:1 › install guide dialog closes on a backdrop click (7.3s)
  ✓  3 test/browser/routes.spec.js:3:1 › route entry renders shell and lists (7.7s)
  ✓  1 test/browser/add-word.spec.js:3:1 › adding a word persists it to the words list (9.0s)

  3 passed (11.6s)
```

```
$ npx shadow-cljs compile app node-test
[:app] Build completed. (331 files, 330 compiled, 0 warnings)
[:node-test] Build completed. (118 files, 117 compiled, 0 warnings)
$ node target/node-tests.js
Ran 144 tests containing 331 assertions.
0 failures, 0 errors.
```

### DIFF
--- a/test/browser/README.md
+++ b/test/browser/README.md
@@ -11,7 +11,7 @@ The suite assumes a backend of its own is already listening on port 8301:
 
 ```sh
 npx shadow-cljs compile app
-LEARNING_APP_PORT=8301 LEARNING_APP_DB_PATH=$PWD/test-app.db \
+LEARNING_APP_PORT=8301 LEARNING_APP__DB_PATH=$PWD/test-app.db \
   clojure -M:dev -m core &
 ```
 
--- a/test/browser/README.md
+++ b/test/browser/README.md
@@ -1,0 +1,43 @@
+# Browser suite (Playwright)
+
+Headless Playwright specs that replace the ad-hoc CDP checks (#169). They run
+inside WSL against the system Google Chrome (`channel: "chrome"`, no browser
+download) — never against the shared dev stand or the visible Windows Chrome;
+that one remains the CDP skill's territory.
+
+## Running
+
+The suite assumes a backend of its own is already listening on port 8301:
+
+```sh
+npx shadow-cljs compile app
+LEARNING_APP_PORT=8301 LEARNING_APP_DB_PATH=$PWD/test-app.db \
+  clojure -M:dev -m core &
+```
+
+Wait for `curl -s http://localhost:8301/` to answer, then:
+
+```sh
+npm run test:browser   # = npx playwright test
+```
+
+Config is `playwright.config.js` at the repo root: baseURL
+`http://localhost:8301`, `trace: "retain-on-failure"` (traces land in
+`test-results/` on failure).
+
+## Conventions
+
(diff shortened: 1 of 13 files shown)


test/browser/README.md
```diff
@@ -0,0 +1,43 @@
+# Browser suite (Playwright)
+
+Headless Playwright specs that replace the ad-hoc CDP checks (#169). They run
+inside WSL against the system Google Chrome (`channel: "chrome"`, no browser
+download) — never against the shared dev stand or the visible Windows Chrome;
+that one remains the CDP skill's territory.
+
+## Running
+
+The suite assumes a backend of its own is already listening on port 8301:
```
u473t8, line 10: Where did this port come from? As far as I know, we usually access the app via sprecha.local or dev.sprecha.de, which involves nginx reversproxying the couch-db and the backend.

u473t8: It is the suite's own private backend, not the shared stand — that is the point of the port. The agents' rule is that worktree jobs never touch the stand (8083 / sprecha.local / your CouchDB data), so the suite boots its own JVM on a free port with its own SQLite, and the backend alone serves the SPA plus API — enough for every UI flow. What nginx adds — /db/ proxying to CouchDB, auth_request — is exactly the sync territory the suite declares out of scope (README states it). localhost is a secure context, so no TLS needed. 8301 itself is arbitrary; when the staging ladder lands, the same specs can point at a full stand via baseURL.

PR #211, opened 2026-08-06
GH-206 CouchDB joins the backup archive

u473t8:
Closes #206.

`backup.sh` archived exactly one artifact — the SQLite snapshot — while every user's vocabulary, reviews and collections live in CouchDB userdbs with **no backup at all**. One borg archive (`app-<now>`) now carries both stores taken back to back, so a restore is coherent by construction: an account row and its userdb come from the same moment (the rule the runbook now spells out — restoring one store without the other manufactures the orphan states #182/#205 catch).

Access: `/var/lib/couchdb` is `0750 couchdb:couchdb`; the backup service gains `SupplementaryGroups=couchdb` — read access scoped to the unit, not granted to the `dbmaintainer` account globally. CouchDB files are append-only, so a live copy is recoverable; noted in the script.

Pairs with #209 (which makes the SQLite half real — today it snapshots the empty path). No hard gate: archiving CouchDB is valuable either way.

Verified here: bash syntax, `systemd-analyze verify` clean. Honest remainder: the next real backup run must show the couch tree in `borg list`; the restore drill is #191's deliverable. Old `db-*.sqlite` archives stay valid, SQLite-only.

### DIFF
--- a/infra/production/usr/share/learning-app/backup.sh
+++ b/infra/production/usr/share/learning-app/backup.sh
@@ -29,13 +29,18 @@ sqlite3 ${LEARNING_APP_DB_BACKUP_PATH} ".selftest"
 
 sqlite_backup_exit=$?
 
-# Create remote backup
+# Create remote backup. One archive carries both stores so a restore is
+# coherent by construction: an account row and its userdb come from the same
+# moment. CouchDB files are append-only, so a live copy is recoverable —
+# still, the archive is taken right after the sqlite snapshot to keep the
+# two as close as possible.
 borg create                     \
     --list                      \
     --stats                     \
     --show-rc                   \
-    ::db-{now}.sqlite           \
-    $LEARNING_APP_DB_BACKUP_PATH
+    ::app-{now}                 \
+    $LEARNING_APP_DB_BACKUP_PATH \
+    /var/lib/couchdb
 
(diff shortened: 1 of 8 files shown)


infra/production/usr/share/learning-app/backup.sh
```diff
@@ -29,13 +29,18 @@ sqlite3 ${LEARNING_APP_DB_BACKUP_PATH} ".selftest"
...
+# two as close as possible.
 borg create                     \
     --list                      \
     --stats                     \
     --show-rc                   \
-    ::db-{now}.sqlite           \
-    $LEARNING_APP_DB_BACKUP_PATH
+    ::app-{now}                 \
+    $LEARNING_APP_DB_BACKUP_PATH \
+    /var/lib/couchdb
```
u473t8, line 43: Is it a file or a directory? Can borg backup directories?

u473t8: A directory — and archiving directories is borg's native mode: it recurses, stores the tree with ownership and permissions, and deduplicates across archives (unchanged couch shards cost almost nothing in the repo). So the one borg create call carries a file (the sqlite snapshot) and a tree (/var/lib/couchdb) side by side; borg extract restores the tree as it was.

PR #212, opened 2026-08-06
GH-178 keep suggestion list steady while typing

u473t8:
Closes #178.

Delivers the "Also worth a look" half only: `:action/update-word` no longer resets `:home/suggestions` per keystroke, so the previous list stays visible through the 150 ms debounce + query instead of flashing empty. Same render count, steadier look.

**Not implemented on purpose:** auto-filling the translation from the top suggestion (the merge of dictionary answer → translation) is a UX decision for the owner and stays open — hence "Part of", not "Closes".

## Suggestions lifecycle — every write to `:home/suggestions` and its fate

| Write site (`pages/home/actions.cljs`) | Fate |
| --- | --- |
| `:action/update-word` | **Changed**: no longer writes suggestions; saves only `:home/word` + cleared `:home/translation` |
| `:action/update-suggestions` | Unchanged: dictionary answer replaces the list, active-idx reset to 0 |
| `:action/show-home` | Unchanged: clears on page load **and** on successful submit (`:effect/add-word` success dispatches it) |
| `:action/dismiss-suggestions` | Unchanged: input blur clears |
| `select-item` (click pick via `:action/select-suggestion`, Enter/Tab via keydown) | Unchanged: pick clears |
| keydown ArrowDown/ArrowUp | Unchanged: rewrites with new active-idx |
| keydown Escape | Unchanged: clears |

Emptied input still clears: `adapters.dictionary/completions` returns `[]` for an empty prefix, `[]` is truthy, so the debounced `suggest!` dispatches `:action/update-suggestions []` and the list empties ~150 ms later through the normal path. Add-error (`:action/show-word-error`) never touched suggestions before and still does not; in that flow the list was already dismissed by blur when focus left the word input.

**Stale list stays coherent on pick:** while the visible list belongs to the previous prefix, picking an entry fills word and translation from the entry's own `:lemma`/`:translations` (`select-item`), and keydown picks from the same state list that is rendered — never a mix with the current input.

Edge worth naming: `suggest!` skips dispatch when the dictionary is not ready, so a kept list would then outlive its prefix — but a visible list implies a prior successful lookup, i.e. the dictionary was ready, and readiness does not regress.

## Tests

`test/client/pages/home_test.cljs` (nexus bootstrap local to the test: state fn, capability interceptor, plain `:effect/save`):

- `second-keystroke-keeps-suggestions`
- `dictionary-answer-replaces-suggestions`
- `emptied-input-clears-suggestions`
- `submit-clears-suggestions`
- `picking-a-kept-suggestion-uses-its-own-data`

## Verification

- `npx shadow-cljs compile app node-test` — 0 warnings
- `node target/node-tests.js` — 149 tests / 343 assertions (baseline 144/331), 0 failures
- Visual steadiness while typing **pends the shared dev stand** — not checked in a browser here
- OpenSpec: requirement added to `home-add-form-focus`, change archived as `2026-08-06-steady-suggestion-list`

### DIFF
--- a/openspec/changes/archive/2026-08-06-steady-suggestion-list/.openspec.yaml
+++ b/openspec/changes/archive/2026-08-06-steady-suggestion-list/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven-with-adr
+created: 2026-08-06
--- a/openspec/changes/archive/2026-08-06-steady-suggestion-list/proposal.md
+++ b/openspec/changes/archive/2026-08-06-steady-suggestion-list/proposal.md
@@ -1,0 +1,18 @@
+# Suggestion list stops flashing between keystrokes
+
+## Why
+
+GH-178 ("Also worth a look"): every keystroke dispatches `:action/update-word`, which resets `:home/suggestions` along with saving the word. The list goes blank for the 150 ms debounce plus the dictionary query on every character, so it visibly flashes while typing. Keeping the previous list until `:action/update-suggestions` delivers the next answer renders the same number of times but looks steady.
+
+Out of scope: auto-filling the translation from the top suggestion (the other half of GH-178) is an open UX question for the owner and is deliberately not implemented here.
+
+## What changes
+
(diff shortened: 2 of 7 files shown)


u473t8 (commented):
LGTM

PR #213, opened 2026-08-06
GH-176 render once per dispatch without effect batching

u473t8:
Closes #176.

## What

- `nexus.batching/install!` removed; `:effect/save` back to the plain unbatched signature.
- Render scheduling: `application/install-render!` owns a dispatch-depth counter and a dirty flag. The store watch renders immediately outside a dispatch (external writes keep rendering) and marks dirty inside one; `:after-dispatch` back at depth zero renders once if dirty. A counter, not a flag — nested dispatches (dialog on-mount) and async continuations (new top-level dispatches) both keep the guard up correctly.
- `window.__metrics()` now counts renders at the render call instead of per store notification: with unbatched saves one dispatch may notify the store several times while rendering once.

## Semantics change: mid-dispatch state reads

Batched `:effect/save` used to execute after every other effect in the dispatch, so a state-reading effect (e.g. `:effect/exit-editing-on-background`) saw pre-save state whenever a save shared its dispatch. Unbatched saves execute in place (nexus ADR-01), so effects and later-expanded actions now see post-save state. `:effect/exit-editing-on-background` itself is unaffected in practice — it is dispatched alone from the collections background click and issues its own saves via nested dispatch — but the reordering hazard #176 names is gone for any future dispatch that combines a save with a state read.

Also: the sync-menu dialog's on-mount dispatch now nests inside the after-dispatch render (depth 1) instead of inside a mid-dispatch render (depth 2); `__metrics` will classify it as a top-level dispatch.

## Verification

- `npx shadow-cljs compile app node-test` — 0 warnings.
- `node target/node-tests.js` — 149 tests, 340 assertions, 0 failures (144 tests before; the 5 new ones are `client.render-test`): several saves render once, no-save dispatch renders nothing, nested dispatch renders once, external `swap!` renders, async continuation renders on its own.
- **Pending:** in-browser render-count confirmation of the #176 table via `window.__metrics()` on the dev stand — not done in this PR.

OpenSpec: `main-thread-runtime` requirement modified, change archived as `2026-08-06-render-once-per-dispatch`.

### DIFF
--- a/src/client/application.cljs
+++ b/src/client/application.cljs
@@ -438,30 +438,19 @@
 ;; A counter, not a flag: effects dispatch actions from inside a dispatch
 ;; (dialog on-mount), and an async continuation arrives as a new top-level
 ;; dispatch — both must keep the guard up until their own dispatch unwinds.
-(defonce ^:private dispatch-depth (volatile! 0))
-
-
-(defonce ^:private dirty? (volatile! false))
-
-
 (defn install-render!
+  "Renders on every state change and only on change. `identical?` is enough:
+   CLJS `assoc`/`merge` hand back the same map when nothing differs, so a
+   dispatch that saves nothing costs no render. A dispatch that saves several
+   times renders several times — measured earlier: extra renders inside one
+   synchronous dispatch cost diffing, never extra paints, and the owner chose
+   this simplicity over a coalescing counter (#213)."
   [store render-fn]
   (add-watch store
              ::render
-             (fn [_ _ _ state]
-               (if (pos? @dispatch-depth)
-                 (vreset! dirty? true)
-                 (render-fn state))))
-  (nxr/register-interceptor!
-    {:before-dispatch (fn [ctx]
-                        (vswap! dispatch-depth inc)
-                        ctx)
-     :after-dispatch  (fn [ctx]
-                        (vswap! dispatch-depth dec)
-                        (when (and (zero? @dispatch-depth) @dirty?)
-                          (vreset! dirty? false)
-                          (render-fn @store))
-                        ctx)}))
+             (fn [_ _ old state]
+               (when-not (identical? old state)
+                 (render-fn state)))))
 
 
 (defn routes
--- a/src/client/application.cljs
+++ b/src/client/application.cljs
@@ -5,7 +5,6 @@
    [application.presenter :as presenter]
    [install-guide.view :as install-guide]
    [lambdaisland.glogi :as log]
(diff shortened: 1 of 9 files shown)


src/client/application.cljs
```diff
@@ -441,6 +431,39 @@
...
+                        (vswap! dispatch-depth inc)
+                        ctx)
+     :after-dispatch  (fn [ctx]
+                        (vswap! dispatch-depth dec)
+                        (when (and (zero? @dispatch-depth) @dirty?)
+                          (vreset! dirty? false)
+                          (render-fn @store))
+                        ctx)}))
+
+
```
u473t8, line 455: I am still not convinced. Did we solve a real problem here? Cannot we just compare two atom states in atom watch to decide to render or not? Will repaint happen, if nested dispatch changes atom value as well as top-level?

u473t8: Three honest answers. (1) What real problem: not renders-per-dispatch CPU — we measured earlier that extra renders inside one dispatch cost diffing, not paints. The drivers are that batching is deprecated upstream (staying on it pins us to old nexus) and that it reordered effects: batched saves ran after everything, so a state-reading effect in the same dispatch saw pre-save state — the loaded gun we flagged in :effect/exit-editing-on-background. The counter is how we removed batching without regressing the render table. (2) Compare-in-watch: it solves a different row of that table. Equality-skip catches 'nothing changed' (and cheaply — CLJS assoc returns the identical map for an identical value), but three saves of three different values still fire the watch three times → three renders. The counter collapses those into one and keeps zero for no-ops; compare alone cannot do the first. (3) Nested dispatch changing the atom: yes, exactly one render — the nested after-dispatch decrements to depth 1 and does not render; the top-level one reaches depth 0, sees dirty, renders once covering both writes. Test nested-dispatch-renders-once pins precisely this, and the browser paints at most once per frame regardless. Leaving this thread open — if you weigh simplicity above the deprecation-and-order argument, compare-only is a legitimate smaller PR and I will reshape.

u473t8: Reshaped as you chose — simplicity wins: the counter and dirty flag are gone, the watch renders on `identical?` change and nothing else. Zero-render dispatches survive (assoc returns the identical map for an identical value — pinned by a new test), several distinct saves render per change (also pinned, as the documented cost), nested and async behavior re-pinned under the new semantics. Batching removal and the effect-order fix stay untouched. Spec reworded in both the main spec and the archived delta. Commit: 7e16fd5.

PR #221, opened 2026-08-06
GH-218 Boot-time telemetry visible under the dev alias

u473t8:
Closes #218.

Boot events (adoption, migrations, reconciliation) logged before any handler existed — under `-M:dev` telemere ships none, so they vanished. Found live: #208's report ran fine but its line never appeared. Handler now installs first in -main. Plus reflection hints (adoption code + two old ones in the jar test).

Verified on the stand: the orphan-userdbs line appears at boot, zero reflection warnings, 22 tests green. Production was unaffected — the packaged run has a default handler.

### DIFF
--- a/src/backend/core.clj
+++ b/src/backend/core.clj
@@ -95,18 +95,15 @@
               target-wal (io/file (str (.getPath target) "-wal"))
               tmp        (io/file (str (.getPath target) ".adopting"))]
           (some-> (.getParentFile target) .mkdirs)
-          (let [replace ^"[Ljava.nio.file.CopyOption;"
-                        (into-array
-                         java.nio.file.CopyOption
-                         [java.nio.file.StandardCopyOption/REPLACE_EXISTING])
-                atomic  ^"[Ljava.nio.file.CopyOption;"
-                        (into-array
-                         java.nio.file.CopyOption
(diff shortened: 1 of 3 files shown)


src/backend/core.clj
```diff
@@ -95,19 +95,18 @@
...
-                                                  [java.nio.file.StandardCopyOption/REPLACE_EXISTING])))
-          (java.nio.file.Files/copy (.toPath legacy)
-                                    (.toPath tmp)
-                                    (into-array java.nio.file.CopyOption
-                                                [java.nio.file.StandardCopyOption/REPLACE_EXISTING]))
-          (java.nio.file.Files/move (.toPath tmp)
-                                    (.toPath target)
-                                    (into-array java.nio.file.CopyOption
-                                                [java.nio.file.StandardCopyOption/ATOMIC_MOVE]))
+          (let [replace ^"[Ljava.nio.file.CopyOption;"
```
u473t8: These typehints are terrifying!

u473t8: Fair — gone in c168372. io/copy plus File.renameTo give the same semantics (overwriting copy; rename(2) is atomic within a directory) and need no hints at all. The terrifying syntax was the cost of the NIO varargs API, so I removed the API instead of decorating it.

PR #222, opened 2026-08-06
GH-215 ADR-0010: SSH deploys and Cloudflare Tunnel for staging

u473t8:
Part of #215. Records the owner's decisions: rungs 2 and 3 complement each other with no ordering; staging deploys over SSH; exposure via Cloudflare Tunnel (no port forwarding, TLS at the edge, optional Access gating). The self-hosted-runner rule stays in the ADR for the day one appears.


u473t8 (commented):
LGTM

PR #224, opened 2026-08-06
GH-214 Staging rung 2: the production shape in a systemd container from the deb

u473t8:
Part of #214.

Rung 2 of ADR-0010: `infra/staging/` boots the actual deb (built byte-for-byte as `infra-checks.yml` builds it) in a systemd container and smokes the production shape. No ports published; everything asserted through `docker exec`.

## Smoke output (real run, this daemon)

```
PASS unit active: nginx.service
PASS unit active: couchdb.service
PASS unit active: learning-app-run.service
PASS unit active: learning-app-restart.path
PASS unit active: learning-app-certbot.timer
PASS unit active: learning-app-backup-db.timer
PASS nginx: https:// answers 200
PASS nginx: http:// redirects 301
PASS app: /auth/check without cookie is 401
PASS nginx: /db/ without cookie is 401
PASS migrations: schema_migrations has rows
FAIL adoption: database-adopted in the journal
FAIL adoption: marker row survived the move
FAIL adoption: legacy file retired
PASS backup: learning-app-backup-db.service runs clean
PASS backup: archive exists
PASS backup: archive carries the sqlite store
PASS backup: archive carries the couchdb store
SMOKE: 3 assertion(s) failed
```

The postinst OpenRouter gate was also exercised: without the key credential the first install refuses to start `learning-app-run`; a second (idempotent) install with the key staged enables it.

## The three FAILs are a real catch, not a script bug

`tmpfiles.d/learning-app.conf` ships `f /var/lib/learning-app/db.sqlite 0660 webapp webapp -` (GH-64 era), so postinst pre-creates an **empty** file at the configured path before the app ever starts. #209's adoption then takes its no-clobber branch and strands the legacy database. Container evidence:

```
ERROR EVENT ... core[91,9] ::legacy-database-left-behind
-rw-r--r-- webapp webapp  8192 /opt/learning-app/app.db      <- marker row still here
-rw-rw---- ...           77824 /var/lib/learning-app/db.sqlite <- fresh empty schema
```

The same empty file exists on production (created by the same tmpfiles line on every install since GH-64), so the not-yet-deployed #209 would strand the accounts on its first real deploy. Exactly the class of bug this rung exists to catch — details commented on #209. Once that lands, this smoke goes green.

## systemd-in-docker incantation (recorded per ADR-0010)

1. PID 1 dies with "Failed to create /init.scope control group" under the default private cgroupns → `--cgroupns=host -v /sys/fs/cgroup:/sys/fs/cgroup:rw --tmpfs /run --tmpfs /run/lock`.
2. Unit sandboxes die with "Failed to keep CAP_SYS_ADMIN" (status=217/USER) → `--cap-add=SYS_ADMIN --security-opt seccomp=unconfined --security-opt apparmor=unconfined`.
3. The subtle one: units start but `/run/credentials/<unit>` is silently empty — docker leaves mounts private and systemd-in-a-container skips its "remount / rshared" boot step, so the credential tmpfs is MS_MOVEd into an unobserved namespace → `mount --make-rshared /` after boot. `--privileged` does **not** fix this and is not needed at all.

`systemd-creds encrypt --with-key=host` works inside the container, so postinst's encrypted-credential flow runs for real.

## Also observed (footnotes, no action taken here)

- `learning-app-run` (User=webapp) and `learning-app-backup-db` (User=dbmaintainer) share `StateDirectory=learning-app`; every service start rechowns the whole state dir to its own user, so `db.sqlite` ownership ping-pongs between backups and app restarts. Self-heals on each start today, but it is accidental.
- `deps.edn`'s `:build` alias pins tools.build `{:git/tag "v0.10.14" :git/sha "3a3c177"}` — that sha is v0.10.13; a clean gitlibs cache fails `clojure -T:build uber` with "sha and tag point to different commits" (CI passes only via cache).
- Our postinst never restarts CouchDB, so the deb's `10-learning-app.ini` (proxy auth handlers) is not loaded until the next CouchDB restart — true on production too.

What is and is not proven by a green run is written down in `infra/staging/README.md` (certbot dry, local borg with a plaintext test passphrase, shared kernel; reality is #215's rung).

### DIFF
--- a/infra/staging/Dockerfile
+++ b/infra/staging/Dockerfile
@@ -1,0 +1,74 @@
+# Staging rung 2 (ADR-0010): the production shape in a systemd container.
+#
+# Base: ubuntu:24.04 — not pure Debian. Production is an Ubuntu Hetzner box
+# (admin-setup.sh --bootstrap uses `add-apt-repository universe`), and noble is
+# the release that carries openjdk-21 and has CouchDB packages in the Apache
+# apt repository. This is the closest image to what the deb actually lands on.
+FROM ubuntu:24.04
+
+ENV DEBIAN_FRONTEND=noninteractive container=docker
+
+# A hung mirror connection must fail fast and retry, not stall the build.
+RUN printf '%s\n' \
+        'Acquire::Retries "3";' \
+        'Acquire::http::Timeout "30";' \
+        'Acquire::https::Timeout "30";' \
+    > /etc/apt/apt.conf.d/80-staging-timeouts
+
+# systemd as PID 1 plus tooling for the repo setup below.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        systemd systemd-sysv dbus \
+        curl ca-certificates gnupg apt-transport-https openssl \
+    && rm -rf /var/lib/apt/lists/*
+
+# CouchDB apt repository — the same source admin-setup.sh --bootstrap adds.
+RUN curl -fsSL https://couchdb.apache.org/repo/keys.asc \
+        | gpg --dearmor -o /usr/share/keyrings/couchdb-archive-keyring.gpg \
+    && . /etc/os-release \
+    && echo "deb [signed-by=/usr/share/keyrings/couchdb-archive-keyring.gpg] https://apache.jfrog.io/artifactory/couchdb-deb/ ${VERSION_CODENAME} main" \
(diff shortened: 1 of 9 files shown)


u473t8 (commented):
Incomprehensible for me, therefore LGTM :) I was expecting integrating checks into CI pipeline, did I get it wrong?

PR #226, opened 2026-08-06
GH-225 Adopt over the tmpfiles-pre-created empty target

u473t8:
Closes #225.

The #214 container smoke caught this before any deploy: the deb's tmpfiles line pre-creates an empty `/var/lib/learning-app/db.sqlite`, #209's no-clobber check treated it as data and refused adoption — the app would start on a fresh empty schema with accounts stranded in `/opt`. Production has the same pre-created file, so current master must not deploy without this.

Fix: a zero-length target is deleted before the check; a non-empty target still refuses (rollback safety intact). New test pins it; 23 backend tests green. The reflection warnings in the run are master's, fixed by the open #221.

Re-running #224's container smoke after merge should flip its 3 FAILs to PASS — that's the end-to-end confirmation.

### DIFF
--- a/src/backend/core.clj
+++ b/src/backend/core.clj
@@ -85,6 +85,11 @@
    leftover is the operator's decision."
   [^java.io.File legacy {:keys [dbname]}]
   (let [target (io/file dbname)]
+    ;; systemd-tmpfiles pre-creates the target as an empty file (#225), and an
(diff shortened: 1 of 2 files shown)


u473t8 (commented):
Do I get it right, that after proper migrations we can drop this legacy handling code?

PR #233, opened 2026-08-06
GH-228 single owner for the shared state directory

u473t8:
Closes #228.

## Choice: drop the backup's claim, don't split state homes

`learning-app-backup-db` declared `StateDirectory=learning-app` for one reason only (its own comment said so): keeping `/var/lib/learning-app` writable under `ProtectSystem=strict`, because sqlite3 needs to touch journal/-wal/-shm next to the source db even for a read. It has no state of its own there — borg cache/config live in `/var/lib/dbmaintainer`, already `ReadWritePaths`. A separate state home would give it a directory it stores nothing in and leave the write-access need unsolved. So:

- **backup-db**: `StateDirectory` and `LogsDirectory` dropped (it logs to the journal only — same rechown mechanism applied to `/var/log/learning-app`); `/var/lib/learning-app` added to `ReadWritePaths`, which punches through `strict` with zero ownership semantics.
- **run**: sole owner, pins `StateDirectoryMode=2770` — setgid group-writable, so dbmaintainer (member of group `webapp` via sysusers) can create sqlite side files when none exist without owning the tree.
- **tmpfiles.d**: `/var/lib/learning-app` 0750 → 2770 to match; postinst's `systemd-tmpfiles --create` converges existing installs.

## Both start orders

- *backup then app start*: backup no longer declares the directory, nothing rechowns; top dir stays webapp:webapp, so the app's `StateDirectory` sees matching ownership and does **not** recurse — a dbmaintainer-owned `-wal` left behind stays group-webapp rw via setgid and remains usable by the app.
- *app start then backup*: dir is webapp:webapp 2770; backup reads `db.sqlite` (0660, group webapp) and creates/uses side files via group write on the directory. Previously this leg worked only because the backup's own start had rechowned the tree to dbmaintainer.

Side effect fixed for free: the app's recursive rechown no longer periodically flips `/var/lib/learning-app/dictionary` away from its intended `deployer:webapp`.

## Verification

- `systemd-analyze verify` on both changed units: clean, exit 0 (`StateDirectoryMode=2770` parses).
- Reasoning above for both start orders; the app-down backup corner (no `-wal` present, dbmaintainer must create it) is covered by the setgid group-writable mode.
- Not proven live here: the real chown behavior end-to-end is exactly what the #214 container smoke (`infra/staging/run.sh`) exercises — that is where this should be confirmed green.

OpenSpec: delta on `production-service-hardening` (new requirement: one owning unit per shared resource directory), validated and archived in the same commit.

### DIFF
--- a/infra/production/etc/systemd/system/learning-app-backup-db.service
+++ b/infra/production/etc/systemd/system/learning-app-backup-db.service
@@ -8,8 +8,6 @@ User=dbmaintainer
 # granted to this service only, not to the user account globally.
 SupplementaryGroups=couchdb
 WorkingDirectory=~
-LogsDirectory=learning-app
-StateDirectory=learning-app
 ExecStart=sh -c './backup.sh'
 
 # Environment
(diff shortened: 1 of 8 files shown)


u473t8 (commented):
LGTM

PR #234, opened 2026-08-06
GH-230 Run the container smoke on infra PRs in CI

u473t8:
Closes #230.

## What

`.github/workflows/staging-smoke.yml`: on every PR to master a cheap `changes` job diffs the PR range; when `infra/**` or the workflow file itself changed, the heavy job builds the jar (the deploy-app.yml way) and runs `infra/staging/run.sh` on ubuntu-latest; otherwise it reports skipped. Same always-reporting shape as integration.yml — a required check that never starts blocks the merge forever (#204). 20-minute timeout; smoke output uploaded as an artifact on failure. No container ports published; the systemd-in-docker incantation from `infra/staging/README.md` (cgroupns=host, rw cgroup bind, CAP_SYS_ADMIN, unconfined seccomp/apparmor) is permitted on GitHub-hosted runners.

**Not done here:** adding `Container smoke` to branch protection's required checks — that is the owner's console click.

**Rider fix (needed for the CI jar build):** `deps.edn` pinned tools.build `:git/tag "v0.10.14"` with v0.10.13's sha `3a3c177`; the current Clojure CLI refuses the mismatch (`sha and tag that point to different commits`) and `clojure -T:build uber` dies — locally and on any fresh runner. Sha corrected to `1176afd`, the commit v0.10.14 actually tags.

## Local smoke against current master (with #226 merged)

Previously 15 PASS / 3 FAIL (all three on adoption — the tmpfiles pre-created empty target, README's "Known red"). Fresh run, fully green:

```
PASS unit active: nginx.service
PASS unit active: couchdb.service
PASS unit active: learning-app-run.service
PASS unit active: learning-app-restart.path
PASS unit active: learning-app-certbot.timer
PASS unit active: learning-app-backup-db.timer
PASS nginx: https:// answers 200
PASS nginx: http:// redirects 301
PASS app: /auth/check without cookie is 401
PASS nginx: /db/ without cookie is 401
PASS migrations: schema_migrations has rows
PASS adoption: database-adopted in the journal
PASS adoption: marker row survived the move
PASS adoption: legacy file retired
PASS backup: learning-app-backup-db.service runs clean
PASS backup: archive exists
PASS backup: archive carries the sqlite store
PASS backup: archive carries the couchdb store
SMOKE: all assertions passed
```

18 PASS, 0 FAIL, exit 0 — evidence for closing #214's "known red" note.

## Proof in Actions

The workflow file is in its own path filter, so this PR must run the smoke itself. Result reported below once checks finish.

### DIFF
--- a/.github/workflows/staging-smoke.yml
+++ b/.github/workflows/staging-smoke.yml
@@ -1,0 +1,96 @@
+name: Staging smoke
+on:
+  pull_request:
+    branches: [master]
+
+jobs:
+  # A required check that never reports blocks the merge forever, and a path
+  # filter on the trigger is exactly how a check never reports (#204). So the
+  # workflow fires on every PR and decides here: a skipped job satisfies
+  # branch protection, a never-started workflow does not.
+  changes:
+    name: Detect infra changes
+    runs-on: ubuntu-latest
+    outputs:
+      relevant: ${{ steps.diff.outputs.relevant }}
+    steps:
+      - uses: actions/checkout@v7.0.1
+        with:
+          fetch-depth: 0
+      - id: diff
+        env:
+          BASE: ${{ github.event.pull_request.base.sha }}
+        run: |
+          if git diff --name-only "$BASE" HEAD | grep -qE '^(infra/|\.github/workflows/staging-smoke\.yml$)'; then
+            echo "relevant=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "relevant=false" >> "$GITHUB_OUTPUT"
+          fi
+
+  container-smoke:
+    name: Container smoke
+    runs-on: ubuntu-latest
+    needs: changes
+    if: ${{ needs.changes.outputs.relevant == 'true' }}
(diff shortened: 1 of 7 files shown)


.github/workflows/staging-smoke.yml
```diff
@@ -0,0 +1,96 @@
...
+        with:
+          node-version: 20
+          cache: npm
+          cache-dependency-path: package-lock.json
+
+      - uses: DeLaGuardo/setup-clojure@13.6.1
+        with:
+          cli: latest
+
+      - id: cache-clojure
```
u473t8, line 56: Does it repeat existing uberjar building CI config?

u473t8: Honest yes — the jar recipe (npm ci → shadow release → -T:build uber) now lives in deploy-app.yml and here. Different purposes (deploy publishes and caches, smoke just needs an artifact to install), but the duplication is real and will drift. Filed as a small node: extract the build into a reusable workflow both call.

PR #235, opened 2026-08-06
GH-229 restart CouchDB in postinst when the shipped ini changed

u473t8:
Closes #229.

## Choice: sha256 stamp, restart only on change

dpkg gives postinst no signal about whether a conffile-like payload actually changed (`/opt/couchdb/etc/local.d/10-learning-app.ini` is a plain payload file, not a dpkg conffile, so there is no `ucf`/conffile machinery to lean on). Simplest honest means: postinst hashes the shipped ini and compares against a stamp from the previous install (`/var/lib/misc/learning-app-couchdb-ini.sha256`); on mismatch it restarts `couchdb.service` and rewrites the stamp.

Properties:

- **Idempotent**: re-running the install with the same ini finds a matching stamp and performs no restart.
- **Ordering preserved**: the restart sits between `systemctl enable --now couchdb.service` and the existing 30-second readiness wait-loop, so the wait and the admin/dictionary-db configuration run against the restarted instance — the postinst flow that #224 exercised still holds.
- **First stamped install restarts once** (stamp missing counts as changed). Deliberate: on hosts where CouchDB predates the stamp — production included — that one restart is exactly what loads the shipped proxy-auth handlers that today are only loaded by accident.
- Stamp update sits after the restart under `set -e`, so a failed restart leaves the stamp unwritten and the next install retries.

## Verification

- `bash -n infra/production/DEBIAN/postinst` clean (shellcheck not available on this machine).
- Honestly: this is syntax plus reasoning only. The behavioral proof — upgrade over a running CouchDB restarts it, second install does not, admin setup still lands — is exactly what the #214 container smoke (`infra/staging/run.sh`) exists for; I did not run it here (the docker daemon is in use by other work).

OpenSpec: delta on `deploy-pipeline` (new requirement: config shipped in the deb is in effect when postinst finishes), validated and archived in the same commit.

### DIFF
--- a/infra/production/DEBIAN/postinst
+++ b/infra/production/DEBIAN/postinst
@@ -61,6 +61,23 @@ fi
 # Enable CouchDB and configure admin
 systemctl enable --now couchdb.service
 
+# CouchDB reads local.d only at startup, so a changed 10-learning-app.ini
+# riding the deb would otherwise apply on the next unrelated restart (GH-229).
+# Restart when the shipped content differs from what the last install stamped;
+# repeat installs with the same ini stay restart-free. The restart sits before
+# the wait-loop below, so readiness and admin setup run against the instance
+# that has the shipped config loaded.
+COUCH_INI="/opt/couchdb/etc/local.d/10-learning-app.ini"
+COUCH_INI_STAMP="/var/lib/misc/learning-app-couchdb-ini.sha256"
+if [ -f "${COUCH_INI}" ]; then
+    ini_hash="$(sha256sum "${COUCH_INI}" | cut -d' ' -f1)"
+    if [ ! -f "${COUCH_INI_STAMP}" ] || [ "$(cat "${COUCH_INI_STAMP}")" != "${ini_hash}" ]; then
+        systemctl restart couchdb.service
+        mkdir -p "$(dirname "${COUCH_INI_STAMP}")"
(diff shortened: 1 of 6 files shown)


infra/production/DEBIAN/postinst
```diff
@@ -61,6 +61,23 @@ fi
 # Enable CouchDB and configure admin
 systemctl enable --now couchdb.service
 
+# CouchDB reads local.d only at startup, so a changed 10-learning-app.ini
+# riding the deb would otherwise apply on the next unrelated restart (GH-229).
+# Restart when the shipped content differs from what the last install stamped;
+# repeat installs with the same ini stay restart-free. The restart sits before
+# the wait-loop below, so readiness and admin setup run against the instance
+# that has the shipped config loaded.
+COUCH_INI="/opt/couchdb/etc/local.d/10-learning-app.ini"
```
u473t8, line 70: Where does this init file come from originally? Do we maintain one?

u473t8: Ours, in the repo: infra/production/opt/couchdb/etc/local.d/10-learning-app.ini — the deb installs it into CouchDB's local.d, which CouchDB reads at start. It binds CouchDB to localhost and wires the proxy-auth handlers for /db/ (born in the sync work, ecb63e3). That provenance is exactly why this PR exists: we ship changes to a file we own, but until now nothing made CouchDB re-read it.

PR #238, opened 2026-08-06
GH-217 phase 2a: jar reads secret under both names

u473t8:
Part of #217. Phases 2b (deb unit rename) and 2c (fallback removal) remain.

## What

`db-auth-secret` now resolves through a pure `configured-db-auth-secret` over an env map:

1. `LEARNING_APP__DB_AUTH_SECRET` (new name) wins,
2. `LEARNING_APP_DB_AUTH_SECRET` (old name) as fallback,
3. checkout well-known value (`running-from-source?` unchanged),
4. else throw — the message now names the NEW variable.

The old-name fallback is transition scaffolding; the in-code comment names the removal condition (phase 2c, after the deb rename deploys). #217 tracks removal.

## Why no infra changes here

`LEARNING_APP_DB_AUTH_SECRET_PATH` and the unit's ExecStart export are deb territory — phase 2b. The jar-side dual read alone makes old-unit + new-jar work in either deploy order: the old unit exports the old name and the jar falls back to it; a renamed unit exports the new name and the jar prefers it.

## Proof

Backend suite green: 28 tests / 56 assertions, zero reflection warnings. New tests cover new-only, old-only, both (new wins), neither+packaged (throw names new var), neither+checkout. No dev stand / docker used.

OpenSpec: `backend-configuration` requirement extended with the dual-read scenarios; validated strict and archived as `2026-08-06-secret-dual-read`.

### DIFF
--- a/src/backend/core.clj
+++ b/src/backend/core.clj
@@ -54,13 +54,24 @@
   (some? (io/resource "core.clj")))
 
 
+(defn- configured-db-auth-secret
+  [env from-source?]
+  ;; GH-217 phase 2a: the unit still exports the old single-underscore name,
+  ;; and unit (deb) and reader (jar) ship in different artifacts — so the jar
+  ;; reads both, preferring the new name. The old-name fallback is transition
+  ;; scaffolding: remove it in phase 2c, once the deb's rename (phase 2b) is
+  ;; deployed. #217 tracks the removal.
+  (or (get env "LEARNING_APP__DB_AUTH_SECRET")
+      (get env "LEARNING_APP_DB_AUTH_SECRET")
+      (when from-source? "secret")
+      (throw (ex-info "LEARNING_APP__DB_AUTH_SECRET is required outside a source checkout" {}))))
+
+
 (def db-auth-secret
   "Signs the proxy-auth headers `/auth/check` hands to CouchDB. A checkout may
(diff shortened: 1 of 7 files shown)


src/backend/core.clj
```diff
@@ -54,13 +54,24 @@
...
+  [env from-source?]
+  ;; GH-217 phase 2a: the unit still exports the old single-underscore name,
+  ;; and unit (deb) and reader (jar) ship in different artifacts — so the jar
+  ;; reads both, preferring the new name. The old-name fallback is transition
+  ;; scaffolding: remove it in phase 2c, once the deb's rename (phase 2b) is
+  ;; deployed. #217 tracks the removal.
+  (or (get env "LEARNING_APP__DB_AUTH_SECRET")
+      (get env "LEARNING_APP_DB_AUTH_SECRET")
+      (when from-source? "secret")
+      (throw (ex-info "LEARNING_APP__DB_AUTH_SECRET is required outside a source checkout" {}))))
```
u473t8, line 67: We should not forget to remove these constructions after we finish cleaning envs

u473t8 (commented):
LGTM

src/backend/core.clj
u473t8: Да — снятие фолбэка это фаза 2c в #217, ишуя реоткрыта и держит это.

PR #239, opened 2026-08-06
GH-237 Extract uberjar build into reusable workflow

u473t8:
Closes #237.

Jar recipe (npm ci → shadow release → `clojure -T:build uber`) lived in both deploy-app.yml and staging-smoke.yml — drift waiting to happen (noticed in #234 review).

- New `.github/workflows/build-jar.yml` (`workflow_call`): recipe + clojure/m2 caching, jar uploaded as `app-jar` artifact, sha256 as workflow output.
- deploy-app.yml: build job calls it; deploy job downloads the artifact. Server-side behavior byte-identical: per-run tmp scp name, `sha256sum -c` against the build-time checksum, atomic `mv`.
- staging-smoke.yml: build called behind the `changes` gate (explicit needs/if — called workflows inherit nothing); smoke downloads instead of rebuilding; `changes` filter now also watches build-jar.yml; #204 always-reporting shape preserved. Smoke timeout 20→15 min since the jar build moved out.
- `infra/staging/run.sh` untouched — it already consumes a prebuilt `target/learning-app.jar`.

Verification: staging-smoke fires on this PR (its file changed) and exercises the exact workflow_call + artifact path. deploy-app is workflow_dispatch-only — its path is YAML-validated and graph-checked but not executed; first real proof is the next deploy.

Note: #237 was closed today with no linked PR and master unchanged — looks like a board misclose; this PR is the actual fix.

### DIFF
--- a/.github/workflows/build-jar.yml
+++ b/.github/workflows/build-jar.yml
@@ -1,0 +1,75 @@
+name: Build App Jar
+
+# The single home of the jar recipe (#237). deploy-app.yml and
+# staging-smoke.yml both call this workflow and consume the artifact it
+# uploads, so the recipe cannot drift between them.
+on:
+  workflow_call:
+    outputs:
+      jar-sha256:
+        description: sha256 of target/learning-app.jar, recorded at build time.
+        value: ${{ jobs.build.outputs.jar-sha256 }}
+
+jobs:
+  build:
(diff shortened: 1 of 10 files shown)


u473t8 (commented):
LGTM

PR #268, opened 2026-08-08
Browser suite runs in CI

u473t8:
Closes #260

Adds a Browser Tests job to the integration workflow: compile the app bundle, boot a backend on 8301, run the Playwright suite against the runner's system Chrome. Follows the always-reporting pattern (#204) — the changes job gains a browser output, irrelevant PRs skip the job and still satisfy branch protection. Traces and the backend log are uploaded as an artifact on failure.


u473t8 (commented):
LGTM

PR #269, opened 2026-08-08
Provisioning commits the row and the userdb together

u473t8:
Closes #254

The account row committed before the CouchDB steps ran, so a failed CouchDB step stranded a row with no userdb and no token holder — dead weight that also blocked the id.

- The couch steps (`db/use` + `db/secure`) now run inside the provisioning transaction, next to the recycled-id guard that already lives there: a failed step throws and the insert rolls back.
- A userdb created by the failed attempt is removed with it (best effort) — otherwise the recycled-id guard would 503 the freed id on every retry.
- Tests: a failed couch step leaves no row and destroys the fresh userdb; the next attempt provisions the same id cleanly.
- OpenSpec: identity-provisioning requirement extended with cross-store atomicity, change archived.


src/backend/core.clj
```diff
@@ -248,12 +253,24 @@
...
+                    (catch Exception e
+                      ;; The rollback frees the id for reuse, and the guard
+                      ;; above would refuse it for as long as a leftover
+                      ;; userdb exists — so the userdb this very call created
+                      ;; (the guard just proved the name was free) goes too.
+                      ;; Best effort: if CouchDB is down, the delete fails
+                      ;; like everything else and reconciliation reports the
+                      ;; orphan.
+                      (try @(db/destroy {:name dbname}) (catch Exception _))
+                      (throw e)))
```
u473t8, line 272: I don't see here transactions semantic for inserting SQLite row and creating Couch user-db

u473t8: The couch steps do run inside the SQLite transaction: `db/use` + `db/secure` sit in the `jdbc/with-transaction` body (core.clj:250-262), so a throw from either rolls the INSERT back. A true cross-store transaction does not exist, so the couch side is compensated instead — the userdb this attempt created is destroyed before rethrowing (core.clj:271). Row and secured userdb exist together or not at all.

PR #270, opened 2026-08-08
Admin CLI: mint an invite from the jar

u473t8:
Closes #262

The jar entry point now takes arguments: no arguments serves as before; `mint-invite` runs adoption and migrations against the configured database, stores a single-use grant, prints the invite URL and exits without starting the server. The URL comes from `LEARNING_APP__PUBLIC_URL` (default canonical origin), token in the fragment. Unknown arguments exit nonzero.

Verified: backend tests (URL composition, dispatcher mints a burnable grant, no server up) and a live run against a scratch database — URL printed, grant row present, process exits on its own.


u473t8 (commented):
LGTM

PR #272, opened 2026-08-08
postinst: set the CouchDB admin password via local.d, not HTTP

u473t8:
Closes #252

The admin-password HTTP PUT authenticated with the old password, so it could only re-set a password it already knew; `|| true` hid the failure whenever the credential and the running instance drifted apart.

Now postinst checks the decrypted credential against the running CouchDB first. On a match it does nothing. On a mismatch it writes `[admins]` with the plaintext into `local.d/99-admin-password.ini` (couchdb:couchdb, 640) and restarts CouchDB, which replaces the plaintext with a hash on startup — no old password needed. If the credential still fails after the restart, postinst aborts loudly; the `|| true` is gone. postrm removes the ini on purge.

Staging proves it: `run.sh` rotates the credstore credential, reinstalls the deb, asserts the rotated password is live and the old one dead, then rotates back; `smoke.sh` asserts the credential authenticates and the written ini holds a hash, not plaintext. The local.d precedence (99- overrides the packaged admins ini) and the hash write-back were also verified by hand in the staging container.


u473t8 (commented):
LGTM

u473t8 (commented):
LGTM

Issue #141, opened 2026-04-22
Manage example generation and custom examples


u473t8:
Users need reliable example generation without duplicate fetch tasks, visible example state, manual retry/generate controls, and the ability to add their own example text for a word. Scope: first fix idempotent example-fetch tasks and retry caps; then add example management UI/state and user-owned examples.

PR #286, opened 2026-08-11
GH-285 User phrases as a first-class learnable entity

### LINKED ISSUE #285: Add user phrases as a first-class learnable entity
A phrase is its own entity: not vocab, not an example of a word — "a word and an example in one". Users add multi-word German expressions with a translation and learn them in lessons.

Key decisions:
- Separate doc type `phrase` (`phrase:<normalized>`) with its own spaced-repetition log; replicated like vocab.
- Phrase translation is stored as a single string (bypasses `parse-translations` splitting).
- Lesson trial requires typing the phrase fully correctly — no self-check; ergonomics via token diff, preserved answer on retry, typography-only forgiveness in grading normalization.
- Multi-line auto-growing inputs for phrase and translation (single-line inputs are the pain point).
- No manual mode toggle: the add form detects word vs phrase from the input (space heuristic with article/`sich`/dictionary-lemma exceptions, suggestion pos overrides), with a tappable chip as the escape hatch.
- Dictionary already ships pos=phrase lemmas; multi-word suggestions get a badge and switch the form into phrase mode. No mass import.
- No LLM example generation for phrases.

UX plan agreed with the owner (auto-detected word/phrase mode on the home form, mixed /words list with a badge, single SRS pool).

u473t8:
Closes #285.

A phrase is a vocabulary document that says so: same `vocab:<normalized>` id (ADR-0008), same review log, `kind: "phrase"` in the document (ADR-0011). A document without a kind is a word, so nothing is migrated. The home form auto-detects word vs phrase (a dictionary lemma equal to the input decides by its pos; otherwise a space means phrase, except after an article or `sich`) and says which through its legend and label, with no control of its own; both inputs are auto-growing textareas. Lessons get a phrase trial: exact typing graded through the new typography-only `normalize-for-grading`, a review per graded attempt as words already write, and the error footer exactly as it already was. No LLM examples for phrases; storage routing, sync and import needed no new branch at all.

One value is one entry: adding a phrase whose text already exists merges the translation into that document and leaves its kind alone. In exchange the kind is correctable — turning a phrase into a word is a field write, which the earlier `phrase:` namespace could never allow.

## Verified

- Unit suite green (180 tests / 426 assertions), release build clean, CI browser suite green.
- Real browser (Chrome via CDP, dev stand), run again after the rework: mode detection and prefill while typing, add → `vocab:auf jeden fall` with `kind: "phrase"` and the translation kept whole through its comma, form cleared and heights released on submit, a phrase merging into an existing word without a second document and without touching its kind, the words list as one vocabulary with no badges, the edit dialog giving a phrase a textarea and a word an input, a phrase trial with the whole translation as prompt, a wrong answer showing plain text with a single «ДАЛЕЕ» and no markup, `AUF JEDEN FALL!` and `wie gehts` both graded correct.
- A `phrase:` document left over from before the rework is simply not listed — the shape older clients degrade to, and the reason no migration is needed.
- Sync: not re-run live (no account provisioned on the stand), and no longer phrase-specific — `sync.cljs` now differs from master by two docstrings and a dropped redundant filter, so phrases ride the vocabulary path already in production.

## Fixes from the owner's review

The review found the add form moving under the cursor and a scrollbar inside a one-line field. Both came from this work:

- `autogrow!` measured `scrollHeight`, which omits the borders a `border-box` height carries, so every field sat 4px short of its content and a one-line translation overflowed. Where `field-sizing: content` works the measuring is now skipped entirely (it was overriding a height the browser had already got right); the inline height is cleared on reset, so an emptied field no longer keeps the height its former content earned.
- The form relaid out from a row to a column on the first space, the chip mounted into a zero-height row, and the centred panel turned every height change into a move of the title. The form is one column at every width, the chip slot is reserved, the panel is anchored to the top.
- The translation field was wiped on every keystroke and refilled 100ms later from the top completion, so its content and height oscillated while typing in the *German* field. It now keeps what it holds until the German field is emptied, and a typed or picked translation is never overwritten.
- That last change then let a stale prefill linger: `hätte` filled «иметь», and typing on to `hätte hätte fahrad kätte` kept it under a phrase that no longer contained the word. While the field is untouched it follows the dictionary's current answer, including an answer with nothing to offer; emptying it by hand hands it back to the dictionary.
- The phrase badge is gone from the words list — the label repeated what the row already carries. Type still reaches the view as a prop (wrapping, which editor the dialog opens), it just no longer prints itself. The row's shape is #290.

The third review pass also dropped a rule: a phrase trial used to write at most one review per lesson, so correcting yourself inside a lesson counted for nothing — and it applied to phrases alone, since a word answered wrongly comes back around and writes a second review already. Now every graded attempt is written, whatever the trial type. How the scheduler should weigh a run of close lapses is an SRS-settings question, not a trial-loop one.

The third review pass replaced the entity's shape. Phrases had their own doc type and id namespace; now they are vocabulary documents with a `kind`. The trade is stated in ADR-0011: the namespace made the kind part of the identity, so a mis-detected kind could never be corrected — and the possibility it bought (one text held as both a word and a phrase) is worth four values in a 158k-lemma dictionary, all interjections. CouchDB leaves no third option: `_id` is its only unique key, Mango indexes are secondary and non-unique, and dedup between devices works precisely because two devices compute the same id. Storage routing, the conflict scan, import and the progress-store queries all go back to speaking one type; the code is 15 lines shorter than before phrases.

The second review pass took four more things out or straight:

- **The lesson footer is untouched by this change.** «ИСПРАВИТЬ» and «ПОЗЖЕ» are gone with the state, action, effect and use case behind them; a wrong phrase answer ends the attempt like a wrong word answer. Getting a second attempt without retyping is worth having and is #295.
- **The token diff is gone from this change.** It was never discussed and could not stay phrase-only: word and example trials grade through `normalize-german` while the diff normalized through `normalize-for-grading`, so a token could read as matched while the grader called it wrong. A wrong phrase answer now shows the typed text and the reference as plain text. #288 designs the markup from scratch, for every trial type.
- **The dictionary now settles the add mode.** Only the article rule spoke for two-token input, so typing `das heißt` in full stored a word while picking the same lemma from the suggestions stored a phrase. A completion equal to the value decides by its pos; the dictionary carries seven such lemmas (`das heißt`, `ein paar`, `ein wenig`, `eine Reihe`, …).
- **The conflict scan lost its type filter**, and then, with the rework, its second range: one `vocab:` pass covers both kinds again.

Also renamed, locally: a lesson trial is generated from `vocab`, the word the database already uses, not from a `word` — which read as "not a phrase" and stopped being true. `:word-id` keeps its name: reviews and examples are stored under it.

Measured after the layout fixes on a 390x844 viewport, typing `a → auf → auf  → auf jeden → auf jeden Fall` and back: the value field, the panel title and the submit button hold at the same coordinates through every keystroke. With the suggestion list open the value field and title still hold; the submit button moves ~184px because the list sits in flow on phone widths — pre-existing behaviour of the list, left for #289 to decide.

`test/browser/add-form-stability.spec.js` locks the desktop half of this in: no scrollbar in a one-line field, the value field unmoved when the mode flips, both fields back to their empty height after a submit. Each assertion was checked against the unfixed code — the first two fail there, and the third is bounded by Chrome sizing the fields through CSS, which the spec says out loud.

| Add form (phrase mode) | Words list | Phrase trial |
|---|---|---|
| add form | words | trial |

The mode chip is gone too. It repeated what the legend and the label already say, and the single direction it offered — phrase to word — is the damaging one: a word is saved through `parse-translations`, which splits «Извини, что я опоздал.» into two entries and asks the backend for an example of a multi-word lemma. A multi-word noun is entered as a word by typing its article (`das Tiny House`), the way the dictionary lists it; a word saved as a phrase costs one generated example, and the kind is a field now, so an editor can correct it (#290). Verified in the browser: the field holds the same coordinates through word → phrase → article, and only the two labels change.

## Not covered here

Safari's `field-sizing` fallback is untested (no Safari on the stand), and the phone layer — in-flow suggestions, the keyboard — needs a mobile Playwright project (#289). Follow-ups from the same review: #288 (answer markup for every trial type), #295 (retype the answer in place), #290 (words list row layout, type moves into the edit dialog heading), and requirements for manual example management added to #141.

Two review questions are settled by the rework rather than left open: the multi-type selector is moot (one type, one query), and the id namespace is decided in ADR-0011. What the rework makes possible and does not build: correcting an entry's kind from the edit dialog — the storage now allows it, the UI belongs with #290.

OpenSpec change `user-phrases` will be archived in this branch after the owner's local check.


src/client/adapters/progress_store.cljs
```diff
@@ -79,11 +80,25 @@
     (nil? (:started-at lesson-state)) (assoc :started-at (now-iso clock))))
 
 
+(defn- ^:async learnable-docs
+  "All vocab and phrase docs. Two queries on purpose: db routing keys on a
+   single :type value, so a $in selector would route nowhere."
```
u473t8: I did not get the explanation, can you prove it here in a comment?

src/client/adapters/progress_store.cljs
```diff
@@ -94,7 +109,7 @@
   [dbs clock
    {:keys [order limit offset search word-ids]
     :or   {order :desc}}]
-  (let [{all-docs :docs} (await (find-all dbs "vocab"))
+  (let [all-docs    (await (learnable-docs dbs))
         docs        (cond->> all-docs
                       (some? word-ids) (filter #((set word-ids) (:_id %))))
```
u473t8, line 106: I forgot, why do we pass and use word-ids for text search?

src/client/domain/answer_diff.cljs
```diff
@@ -0,0 +1,95 @@
+(ns domain.answer-diff
```
u473t8: We need to rethink and redesign the whole diff mechanism. It should also be applied to examples. I have an idea but the whole mechanic should be done as a separate issue

src/client/domain/lesson.cljs
```diff
@@ -29,9 +32,16 @@
   (= (:type trial) trial-type-word))
 
 
+(defn phrase-trial?
+  [trial]
+  (= (:type trial) trial-type-phrase))
+
+
 (defn- word->trial
   [word]
```
u473t8: so, the name "word" is misleading here. It would be better to come up with a more common name for words and phrases

src/client/domain/lesson.cljs
```diff
@@ -110,17 +120,40 @@
...
+    (utils/normalize-for-grading (or answer ""))
+    (normalized-answer answer)))
+
+
 (defn remove-trial
   [remaining-trials trial]
   (vec (remove #(= % trial) remaining-trials)))
 
 
+(defn phrase-review-due?
```
u473t8: I do not get, why from the lesson's perspective words and phrases are different?

src/client/domain/lesson.cljs
```diff
@@ -110,17 +120,40 @@
   (utils/normalize-german (or answer "")))
 
 
+(defn- trial-normalized
+  "Phrase trials grade through the typography-forgiving normalization; word
+   and example trials keep the id-grade one."
+  [trial answer]
+  (if (phrase-trial? trial)
```
u473t8, line 129: I do not get, why from the lesson's perspective phrases and other trials are different?

src/client/domain/phrase.cljs
```diff
@@ -0,0 +1,81 @@
...
+
+
+(def ^:private word-pair-prefixes
+  "First tokens after which a two-token value still reads as a single word:
+   articles (der Tisch) and the reflexive marker (sich freuen)."
+  #{"das" "der" "die" "ein" "eine" "sich"})
+
+
+(defn- word-lemma-match?
+  [value {:keys [lemma pos]}]
```
u473t8: I do not understand what is this function for?

src/client/domain/phrase.cljs
```diff
@@ -0,0 +1,81 @@
...
+                   (word-pair-prefixes (str/lower-case (first tokens)))))
+         (not (some #(word-lemma-match? value %) completions)))))
+
+
+(defn phrase-suggestion?
+  "True for a completion that represents a dictionary phrase: pos=phrase and
+   multi-word. Single-word phrase lemmas (hallo, hey) stay words."
+  [{:keys [lemma pos]}]
+  (and (= "phrase" pos)
+       (str/includes? (str lemma) " ")))
```
u473t8, line 83: а второе условие в and тут зачем? У нас что, есть фразы из одного слова?

src/client/domain/phrase.cljs
```diff
@@ -0,0 +1,81 @@
...
+   decides against the completions on screen."
+  [value completions override]
+  (or override
+      (if (phrase-value? value completions) :phrase :word)))
+
+
+(defn update-phrase
+  "Replaces the phrase translation with one whole entry — never split."
+  [doc translation]
+  (assoc doc :translation [(translation-entry translation)]))
```
u473t8, line 97: Вообще этот неймспейс создал странное ощущение, как будто мы устраиваем себе проблемы на ровном  месте.

src/client/pages/home/actions.cljs
```diff
@@ -59,16 +67,22 @@
...
-    (let [{:keys [translations]} (first completions)]
-      [[:effect/save
-        (cond-> {:home/suggestions (suggestions completions)}
-          (and (seq translations) (str/blank? (:home/translation state)))
-          (assoc :home/translation (str/join ", " translations)))]])))
+  (fn update-suggestions [state {:keys [completions value]}]
+    ;; Guarded by the value the query was made for: a debounced answer landing
+    ;; after more typing (say, the pre-space prefix of a phrase) must neither
+    ;; show a stale list nor prefill the translation with a stale word.
+    (when (= value (:home/word state))
```
u473t8, line 74: Поясни, я не понял как этот guard решает проблему

src/client/pages/home/actions.cljs
```diff
@@ -86,19 +100,51 @@
...
+    ;; German field drops it, but it is no longer wiped on every keystroke —
+    ;; that made the field blink and resize while typing in this one.
+    [[:effect/save
+      (cond-> {:home/word value}
+        (and (str/blank? value) (not (:home/translation-typed? state)))
+        (assoc :home/translation ""))]
      [:effect/suggest-completions value]]))
 
 
+(nxr/register-action! :action/toggle-add-mode
```
u473t8: Этого я тоже не понял, какой реальный пример проблемы, которую решает этот toggle

openspec/changes/user-phrases/specs/user-phrases/spec.md
```diff
@@ -0,0 +1,99 @@
...
+#### Scenario: Typographic variants pass
+- **WHEN** the reference is "Wie geht's?" and the user types "wie gehts"
+- **THEN** the answer is graded correct
+
+#### Scenario: Wrong word fails
+- **WHEN** the user replaces a word of the phrase with a different word
+- **THEN** the answer is graded incorrect
+
+### Requirement: Phrase errors show a token diff and preserve the answer
+On an incorrect phrase answer, the error state SHALL highlight non-matching tokens in the user's answer (wrong/extra) and in the reference (missing), computed over grading-normalized tokens. A primary retry action SHALL return to the input with the previous answer preserved; a secondary action SHALL move on (trial stays in the pool).
```
u473t8: Did we discuss retry action? Maybe it is useful, prove it in this case.

openspec/changes/user-phrases/specs/user-phrases/spec.md
```diff
@@ -0,0 +1,99 @@
+# user-phrases Specification
+
+## ADDED Requirements
+
+### Requirement: Phrases are added from the home form with automatic mode detection
+The home add form SHALL detect whether the input is a word or a phrase without a manual toggle. Input containing a space (after trim) SHALL be treated as a phrase, EXCEPT when it is an article (der/die/das/ein/eine) followed by a single token, `sich` followed by a single token, or it matches a non-phrase dictionary lemma from already-fetched completions. Selecting an autocomplete suggestion SHALL set the mode from the suggestion: multi-word pos=phrase suggestions select phrase mode, all others word mode. A tappable chip SHALL appear when phrase mode is detected and SHALL let the user override the detected mode until the form is cleared.
```
u473t8: А какой смысл превращать фразу в слово?

src/client/sync.cljs
```diff
@@ -45,29 +45,40 @@
...
+  "Resolves every conflicted vocab and phrase doc a replication pass left
+   behind."
   [user-db]
   (try
-    (let [{rows :rows} (await (db/all-docs user-db vocabulary-with-conflicts))
-          conflicted   (->> rows
-                            (map :doc)
-                            (filter #(and (-> % :type #{"vocab"}) (-> % :_conflicts seq))))]
+    (let [scanned    (await (all! (map #(db/all-docs user-db (:query %)) conflict-scans)))
+          conflicted (->> (mapcat (fn [{:keys [doc-type]} {rows :rows}]
```
u473t8: Explain increased complexity of the function

src/client/application.cljs
```diff
@@ -94,6 +94,41 @@
...
+      (= "content")))
+
+
+(defn- autogrow!
+  "Fits a textarea's height to its content: collapse, then measure. The
+   fallback for Safari, which lacks `field-sizing: content` — where the
+   property works, measuring only overrides a height the browser got right.
+   `scrollHeight` omits the borders a border-box height must carry."
+  [node]
+  (when (and node (not (css-sized? node)))
```
u473t8, line 114: I am curious how collapsing then setting sizes work in js execution model? Is it done between paints? Or browser first must paint collapsed, then fixed sizes?

src/client/application.cljs
```diff
@@ -94,6 +94,41 @@
       (.setSelectionRange node len len))))
 
 
+(defn- css-sized?
+  "Whether the browser sizes this textarea itself through `field-sizing`.
```
u473t8, line 98: А зачем мы используем такую функцию, у нас что, возможны вариации?

src/client/adapters/progress_store.cljs
u473t8: The proof is in the router, not in PouchDB. `db.pouch/find-all` picks the database from the selector:

```clojure
;; db/pouch.cljs:104
(defn find-all [dbs query]
  (db/find-all (db-for dbs (get-in query [:selector :type])) query))

;; db/pouch.cljs:31
(defn db-for [dbs doc-type]
  (some-> doc-type doc-type->db dbs))   ; doc-type->db is keyed by strings
```

With `:type {:$in ["vocab" "phrase"]}` the lookup key is a map, `doc-type->db` has no such key, `db-for` returns nil, and `db/find-all` then calls `(info nil)` — it throws before a selector ever reaches PouchDB. So it is not "the query returns nothing", it is "the call has no database to run against".

Both types do live in `:user/db`, so one query is possible in principle — it needs the router to learn about multi-type queries, which changes the contract of every caller, not this function. Worth doing if we keep adding learnable types.

src/client/adapters/progress_store.cljs
u473t8: `word-ids` is not part of the text search — it is the collection filter, and it predates this PR (GH-163, the collections feature). `use_cases/vocabulary.cljs:64-69` reads the active collection's `:word-ids` and passes them down, so the list shows only that collection's members; `search` filters independently a few lines below. They just share one function.

src/client/domain/answer_diff.cljs
u473t8: Agreed — that is #288, opened from your review and blocked by this PR. It carries both halves: first honest markup (tokens rendered raw, `:wrong` only for a 1:1 gap, `:extra` only when the reference side of the gap is empty, no pairing claimed at n:m), then extending the diff to word and example trials over a single grading normalization — today examples grade through `normalize-german` while the diff normalizes through `normalize-for-grading`, so a token can look matched while the grader called it wrong.

src/client/domain/lesson.cljs
u473t8: Fair — this function now takes both, and `word` reads as "not a phrase" here.

The catch is that `:word-id` is a persisted field on every review doc and the key `examples` are fetched by, so a full rename is a data migration, not a rename. Local names (`word->trial`, `word-trial?`, `words` args) can change today for free.

I need a noun from you: `learnable` (already used in `progress-store/learnable-docs`), `entry`, `item`, or something you prefer. Tell me which and I will do the local half in this PR and file the persisted half separately.

src/client/domain/lesson.cljs
u473t8: They are not different in kind, and the asymmetry here is not principled — you are right to poke it.

What is true: only phrase trials have «ИСПРАВИТЬ», so only they can be graded twice within seconds, which is why the discipline was written for them. But a word trial answered wrong also stays in `remaining-trials` (`check-answer` removes a trial only when correct), comes back around later in the same lesson, and writes a second review then. Same double-count, just slower to hit.

So the honest rule is "one review per trial per lesson" for every trial type. That changes the review stream for words, which is already shipped data and feeds retention, so I did not fold it into this PR unasked. Say the word and I either generalise it here or file it.

src/client/domain/lesson.cljs
u473t8: Because words and examples grade through `normalize-german` — the same normalization that bakes the id (ADR-0008) — and a phrase needs forgiveness that must never touch ids: apostrophe variants dropped, unicode quotes and dashes flattened, so "Wie geht's?" and "wie gehts" are the same answer.

Two normalizations is the wrong end state, and that is #288: one grading normalization for all three trial types. It stayed out of this PR because it changes grading for words and examples that are already in use.

src/client/domain/phrase.cljs
u473t8: It keeps multi-word dictionary words from being stored as phrases.

The article/`sich` rule above it only covers a two-token `der Tisch` shape. The dictionary holds 1237 multi-word lemmas whose pos is not `phrase` and which that rule does not cover:

```
das Tiny House [noun]      das öffentliche Telefon [noun]
oh ja [intj]               weiß gedeckt [adj]
der Bachelor of Science [noun]
```

Without this check, typing one of those and picking it from the suggestions creates `phrase:<...>` instead of `vocab:<...>` — and the type is baked into the id, so the only fix afterwards is delete and re-add.

src/client/domain/phrase.cljs
u473t8: Есть, 39 штук — в словаре лежат леммы с `pos = 'phrase'` и без пробела:

```
howdy, bye-bye, faarwel, Hammersbald, Weidmannsdank,
Treptow-Köpenick, Steglitz-Zehlendorf, Acker-Gauchheil …
```

То есть частью междометия, частью просто криво размеченные топонимы. Без второго условия выбор `howdy` из подсказок переключал бы форму в режим фразы и сохранял однословный документ `phrase:howdy` — а тип запечён в `_id`, потом только удалять и заводить заново.

src/client/domain/phrase.cljs
u473t8: Ощущение честное, давай разложу, за что мы платим.

Отдельный префикс `phrase:` даёт две вещи: слово и фраза с одинаковым текстом сосуществуют, и тип документа читается из `_id`. Второе — не украшение: на нём стоит и маршрутизация баз (`doc-type->db`), и скан конфликтов диапазоном ключей в `sync.cljs`. Платим ровно тем, что ты видишь: объединяющие запросы в progress-store, второй скан конфликтов, второй find-by-value, половина этого неймспейса.

Альтернатива — документ `vocab:` с флагом `:phrase? true`. Тогда исчезают объединяющие запросы, второй скан и добрая часть `phrase.cljs`; остаются только доменные правила (перевод не резать по запятым, испытание с точным набором). Теряем возможность иметь `hätte` и словом, и фразой, и везде, где тип берётся из id, придётся читать тело документа.

Я бы взял этот размен, если мы согласны, что кросс-типовые дубликаты нам не нужны — предупреждение про них ты и так велел убрать. Пока у фраз нет накопленных данных, переезд дешёвый. Скажи, и заведу отдельной задачей (или переделаю здесь, если хочешь до мержа).

src/client/pages/home/actions.cljs
u473t8: Это гонка, а не «на всякий случай». Запрос к словарю дебаунсится на 100 мс (`home/effects.cljs:28`) и отвечает асинхронно, поэтому ответ приходит уже к другому значению поля.

Сценарий: набираешь `auf` → уходит запрос за «auf». Не дожидаясь ответа, дописываешь до `auf jeden` → форма уже в режиме фразы. Прилетает ответ на «auf»: без guard он покажет список подсказок для «auf» и подставит его перевод в поле — под фразой, к которой это слово отношения не имеет.

Это ровно тот баг, который ты прислал скриншотом («завис перевод от первого слова»), в его гоночной форме. Guard сравнивает значение, для которого делался запрос, с текущим, и просто выкидывает опоздавший ответ.

src/client/pages/home/actions.cljs
u473t8: Детект — эвристика, и она ошибается в обе стороны.

- `Tiny House` без артикля: в словаре лемма `das Tiny House`, а введённое значение с ней не совпадёт, значит режим — фраза, хотя это существительное.
- `sich freuen`: правило про `sich` держит режим слова, а пользователь может захотеть учить именно коллокацию фразой.
- Любое слово, которого просто нет в словаре (имя собственное, свежий термин) и в котором есть пробел.

Цена ошибки не косметическая: тип запечён в `_id`, поэтому неверно угаданный режим = документ не того типа, который чинится только удалением и повторным добавлением. Чип — аварийный выход из этого; в #285 он так и записан («tappable chip as the escape hatch»).

Отдельно: сам чип в форме мы не обсуждали до конца — если он тебе мешает, вариант «переключатель в заголовке панели» остаётся на столе.

openspec/changes/user-phrases/specs/user-phrases/spec.md
u473t8: Yes — it comes from #285 itself: "Lesson trial requires typing the phrase fully correctly — no self-check; ergonomics via token diff, **preserved answer on retry**, typography-only forgiveness in grading normalization."

The case it pays for: a phrase is long, and without retry one wrong letter costs the whole typed answer — you retype «Ich habe keine Ahnung, wovon du sprichst» from scratch. Retry is also free of SRS consequences, since the first graded attempt is the one that writes the review.

If it does not earn its place in real use it is cheap to drop: one button plus `lesson/retry!`.

openspec/changes/user-phrases/specs/user-phrases/spec.md
u473t8: Ничего существующего чип не превращает — он решает, документ какого типа создаст **следующий** submit. Это важно, потому что тип запечён в `_id` и потом не меняется.

Направление «фраза → слово» нужно для многословного значения, которое на деле одно слово, а словарь его не подсказал: `Tiny House` без артикля, имя собственное, свежий термин. Без чипа такое молча уедет в `phrase:`.

Подробнее — в ответе на `home/actions.cljs:114`.

src/client/sync.cljs
u473t8: Fair hit. The whole zip of `doc-type` with results exists only to keep the old `:type` filter alive — and that filter was already belt-and-braces: an id inside `vocab:` … `vocab:￰` can only belong to a vocab doc, because ids are content-addressed with their type prefix. That is the very property the range scan relies on to be correct in the first place.

Dropping it makes the function two queries and one flat filter, simpler than what was there before phrases:

```clojure
(let [scanned    (await (all! (map #(db/all-docs user-db %) conflict-scans)))
      conflicted (->> scanned
                      (mapcat :rows)
                      (map :doc)
                      (filter #(-> % :_conflicts seq)))]
  ...)
```

Pushing that.

src/client/application.cljs
u473t8: No intermediate paint — the collapsed state is never on screen.

Both writes and both reads happen inside one task. Reading `offsetHeight`/`scrollHeight` right after writing `height: auto` forces a **synchronous style and layout recalculation** — that is exactly why it is written this way, otherwise `scrollHeight` would answer from the layout before the collapse. But painting is not part of that: it happens after the task, in the rendering step of the frame, and by then the height already holds the final value.

So the cost is a forced synchronous layout (one per keystroke, on one element), not a flicker. And where `field-sizing: content` works we do not pay even that — `css-sized?` skips the whole thing.

src/client/application.cljs
u473t8: Возможны: Chrome сам тянет textarea по контенту через `field-sizing: content`, за пределами Chromium поддержки нет — проверить на Safari я не могу, стенда с ним нет. То есть JS-путь — это ветка для браузеров без свойства (в первую очередь Safari, твой телефон), а на Chrome измерение только затирало бы высоту, которую браузер уже посчитал правильно. Именно это затирание и рисовало полосу прокрутки в однострочном поле.

Читается строкой намеренно: `(.-fieldSizing style)` под `:advanced` переименуется, свойство станет `undefined`, и релизная сборка молча уедет на путь измерения везде, включая Chrome. Строка `(.getPropertyValue style "field-sizing")` компилятору не по зубам.

src/client/adapters/progress_store.cljs
u473t8: Can we then make find all smarter?

src/client/adapters/progress_store.cljs
u473t8: Я запутался, а  почему у нас это вообще в progress store?

src/client/domain/answer_diff.cljs
u473t8: Can we then drop diff from the spec and from this pr?

src/client/domain/lesson.cljs
u473t8: Тогда вместо кнопки исправить нужно разрешить перепечатывать поле ввода, и при изменение содержимого менять футер на ПРОВЕРИТЬ. Возможно, в другом ПР

src/client/domain/phrase.cljs
u473t8: Я просто не могу представить себе хорошего примера кросс-типового дубликата

src/client/pages/home/actions.cljs
u473t8: Я подумал, а нужно ли нам вообще дебаунсить?

u473t8: Ну так а потом сразу прилетит перевод подсказки к auf jeden, разве нет?

src/client/pages/home/actions.cljs
u473t8: Я вот теперь думаю, а не ошибаемся ли мы, что не заводим типы документов word и phrase?

src/client/application.cljs
u473t8: .-fieldSizing style) под :advanced переименуется, -- а что это за приколы, я не в курсе

src/client/domain/lesson.cljs
u473t8: This is a huge migration of the code base

src/client/adapters/progress_store.cljs
u473t8: Yes, and cheaper than it looks. The router only has to accept a set of types and insist they share one database:

```clojure
(defn- query-db [dbs query]
  (let [t       (get-in query [:selector :type])
        types   (if (map? t) (:$in t) [t])
        targets (into #{} (map doc-type->db) types)]
    (if (= 1 (count targets))
      (dbs (first targets))
      (throw (ex-info "query spans databases" {:types types})))))
```

Then `list-words` keeps one `find-all` with `:type {:$in ["vocab" "phrase"]}` and `learnable-docs` disappears.

The usual argument against `$in` is the index — but we have none here: the only `create-index` in the app is the task index (`tasks.cljs:64`), so a vocab query already runs as a full scan through PouchDB's fallback. `$in` is the same scan, once instead of twice.

This PR or its own?

src/client/adapters/progress_store.cljs
u473t8: Потому что в этой же функции считаются поиск, сортировка и пагинация. Отфильтруй по коллекции выше — и `offset`/`limit` начнут резать не тот набор. Приехало вместе с коллекциями (GH-163).

Правильное место — объект запроса на порту (`{:collection-id … :search … :page …}`), чтобы адаптер вообще не знал слова «word-ids». Это часть отложенного рефакторинга контракта портов, где лежит и утечка `:_id` в презентеры. Завести задачу?

src/client/domain/answer_diff.cljs
u473t8: Yes — and it shrinks the PR in the right place. Doing it now.

Going out: `domain/answer_diff.cljs` and its tests, `diff-segments` and the `diff-*` props in the lesson presenter, `diff-body` in the view, the `.lesson__diff-token*` rules, and the diff requirement in the spec.

Staying: on a wrong phrase answer the error footer shows the correct answer with «ИСПРАВИТЬ» and «ПОЗЖЕ» — the shape word trials already have. So the feature ships without a mechanism we never agreed on, and #288 designs it from scratch for all three trial types.

src/client/domain/lesson.cljs
u473t8: Нравится больше, чем текущий вариант: поле остаётся живым, футер по изменению текста превращается в «ПРОВЕРИТЬ», и отдельный шаг «ИСПРАВИТЬ» становится не нужен.

Побочный эффект в нашу пользу: тогда «одна запись review на испытание за урок» перестаёт быть спецправилом для фраз и становится просто тем, как устроен урок.

Завожу одной задачей вместе с обобщением review-дисциплины — обе про одно и то же место. В этом PR оставляю как есть.

src/client/domain/phrase.cljs
u473t8: Я посчитал по словарю. Из 158 294 лемм (5 206 из них с `pos = phrase`) значений, существующих одновременно и как фраза, и как не-фраза, ровно **четыре**:

```
als ob, hi, hoi, willkommen
```

Междометия и один союз — там, где разделение на два типа пользователю не даёт ничего.

Со стороны пользователя примера у меня тоже нет: чтобы получился дубликат, надо ввести буквально ту же строку и захотеть две карточки с разными переводами и раздельным прогрессом.

То есть возможность, ради которой мы держим отдельный неймспейс, стоит нам union-запросов, второго скана конфликтов и половины `phrase.cljs` — а применений у неё, похоже, нет. Считаю это доводом за `vocab:` + флаг.

src/client/pages/home/actions.cljs
u473t8: Дебаунс тут не про сеть — словарь локальный. Он про то, что каждый символ иначе запускает префиксный поиск по 158k лемм: при восьми знаках в секунду это восемь запросов, из которых семь никто не увидит.

Дёшево это или нет — вопрос замера, а не мнения. Могу снять время одного запроса на стенде и решить по цифре. Если запрос дешёвый, дебаунс правда лишний — guard по значению нужен независимо от него.

u473t8: Прилетит — и это правильный ответ, он и должен победить. Guard защищает от другого: порядок прихода не гарантирован.

Два запроса живут параллельно, и ответ на «auf» может прийти **после** ответа на «auf jeden». Тогда последним в поле окажется перевод старого слова — и там и останется, потому что больше ничего не прилетит. Guard делает правило «побеждает ответ на текущее значение» вместо «побеждает тот, кто приехал последним».

Для фразы, кстати, свежий ответ обычно пустой — подсказок на «auf jeden» нет, — поэтому поле корректно очищается.

src/client/pages/home/actions.cljs
u473t8: Тут я не понял до конца, а гадать не хочу — типы `vocab` и `phrase` у нас как раз есть. Ты про что из трёх:

1. переименовать `vocab` → `word`, чтобы пара читалась симметрично;
2. сделать тип явным выбором пользователя (две кнопки/два поля ввода) вместо эвристики с чипом;
3. наоборот — один тип документа с полем-подтипом (это та самая развилка `vocab:` + флаг).

От ответа зависит, что делать с неймспейсом, так что скажи — и разложу последствия по выбранному.

src/client/application.cljs
u473t8: Closure Compiler в режиме `:advanced` переименовывает свойства, чтобы ужать код: `(.-fieldSizing style)` вполне может стать `style.Ab`. Не трогает он только то, что описано в externs — файлах с объявлениями браузерного API. shadow-cljs выводит externs сам по типам, но `field-sizing` — свойство новое, и в комплекте externs, с которым собирается Closure, его может не быть.

Чем это кончается: в дев-сборке всё работает, в релизной `style.Ab` — `undefined`, `css-sized?` всегда false, измерение включается везде, включая Chrome. То есть ровно та полоса прокрутки, которую ты нашёл, только в проде и без предупреждений при сборке.

Строка `(.getPropertyValue style "field-sizing")` этому не подвержена: имя метода в externs есть точно, а строковый аргумент не переименовывается никогда. Поэтому и оговорка в комментарии — иначе кто-нибудь (в том числе я) «упростит» это обратно на точечный доступ.

src/client/domain/lesson.cljs
u473t8: Согласен — поэтому делаю только локальную половину: `word->trial` → `learnable->trial` и имена аргументов в `domain/lesson.cljs` и `use_cases/lesson.cljs`. Десяток строк, данные не трогаются.

Огромная миграция — это вторая половина: `:word-id` в каждом review-документе, `:word-ids` в коллекциях, `list-words`/`find-word-by-value` на порту. Её не планирую: имя в базе спокойно остаётся историческим, как это обычно и бывает.

src/client/adapters/progress_store.cljs
u473t8: Is there a way and sense to have index for learnable docs which can be either word or phrase?

src/client/adapters/progress_store.cljs
u473t8: Сейчас решим. Но я не помню логику, почему эта фильтрация относится к прогрессу? Я думал прогресс у нас универсальный

src/client/domain/phrase.cljs
u473t8: Согласен, если что, потом заведём это как отдельную задачу, если возникнут проблемы

src/client/pages/home/actions.cljs
u473t8: Я скорее думаю про третье

src/client/application.cljs
u473t8: shadow-cljs выводит externs сам по типам, -- откуда эти данные, из какой части документации? И странно, что compiler не знает про css свойства. Я думал это одна система. Но аргументы принял

src/client/adapters/progress_store.cljs
u473t8: Two answers, and the second is the one I'd take.

A Mango index would work: `create-index` over `[:type]`, and `$in` uses it (one range per value). It costs a design doc, a full build the first time and an incremental update on every write — for a query that runs when the words page opens.

But we already have the index: `_id`. Ids are content-addressed with the type prefix, so `vocab:` … `vocab:￰` is a range read straight off the primary btree — nothing to declare, build or maintain. `sync.cljs` already scans conflicts exactly that way. `list-words` can drop Mango entirely:

```clojure
(defn- ^:async learnable-docs
  [dbs]
  (let [user-db (dbs/db-for dbs "vocab")
        read!   (fn [prefix]
                  (db/all-docs user-db {:endkey       (str prefix ":￰")
                                        :include-docs true
                                        :startkey     (str prefix ":")}))
        scanned (await (js/Promise.all #js [(read! "vocab") (read! "phrase")]))]
    (into [] (comp (mapcat :rows) (map :doc)) scanned)))
```

Which also settles the question next door: for these two types we don't need a smarter `find`, we need to stop calling `find`.

The catch is that it holds only while the type lives in the id. Go with your third option — one `vocab:` type with a subtype flag — and a single range read covers both, so the whole question disappears.

This PR or its own?

src/client/adapters/progress_store.cljs
u473t8: Прогресс универсальный, а фильтр здесь не потому, что он «про прогресс».

Список сортируется по удержанию: `list-words` берёт документы, дотягивает к каждому уровень из review-логов и сортирует по нему, и только потом режет страницу. Фильтр по коллекции обязан случиться до этого соединения — иначе и сортировка, и `offset`/`limit` посчитаются не по тому набору.

Отдельно сбивает имя: `progress-store` — это адаптер над user/device базами целиком (vocab, phrase, review, lesson, collection), а не хранилище одного прогресса. Отсюда и ощущение, что в него попало лишнее: попало не лишнее, просто имя обещает меньше, чем неймспейс делает.

Развязка — объект запроса на порту: адаптер получает `{:collection-id … :search … :page …}` и не знает слова «word-ids». Завести задачу?

src/client/domain/phrase.cljs
u473t8: Принято.

Одна оговорка про сроки, чтобы она не потерялась: цена переезда растёт вместе с числом накопленных `phrase:` документов. Тип запечён в `_id`, поэтому слияние — это пересоздание документов и перенос лога повторений на новые id. Пока фраз почти нет, это скрипт на десяток строк и одна миграция; «когда возникнут проблемы» проблемой станет в том числе сама миграция.

Так что если склоняешься к слиянию (см. ветку про `home/actions.cljs:114`), дешевле решить это до того, как фразами начнут пользоваться, а не после.

src/client/pages/home/actions.cljs
u473t8: Тогда разложу третье: один тип документа, `vocab:<normalized>` с полем-подтипом (`:phrase? true` или `:kind "phrase"`).

Что уходит:

- `learnable-docs` и объединяющие запросы в progress-store — один диапазон `vocab:` покрывает всё;
- второй скан конфликтов в `sync.cljs`;
- `find-phrase-by-value`, ветвление в `:effect/add-word`, регистрация типа в `doc-type->db`;
- вопрос про индекс из соседней ветки отпадает сам собой.

Что появляется:

- тип больше не читается из `_id` — все места, где мы смотрим на префикс, начинают зависеть от загруженного документа;
- `hätte` не может быть одновременно словом и фразой (по словарю таких значений четыре, так что потеря измеримая и маленькая);
- миграция уже созданных `phrase:` документов: новые id, перенос review-лога, членства в коллекциях.

Что не меняется: перевод фразы не режется по запятым, испытание с точным набором, отсутствие LLM-примеров. Эти правила живут на флаге ровно так же, как на неймспейсе — они доменные, а не структурные.

Заводить задачу сейчас или после мержа этого PR?

src/client/application.cljs
u473t8: Проверил вместо того, чтобы пересказывать — и мой аргумент оказался неверным.

**Про externs.** У Closure нет браузера: он знает только то, что написано в его собственных файлах объявлений. Они лежат прямо в jar компилятора — `externs.zip`, внутри `browser/w3c_css.js`, 3785 строк вручную поддерживаемых объявлений вида

```js
/** @type {string} */ CSSStyleDeclaration.prototype.rowGap;
```

В версии, которой собираемся мы (v20250407), `rowGap` есть, `fieldSizing` — нет. Так что «одна система» — это и есть источник странности: файлы пишутся руками и отстают от платформы.

**Про то, в чём я ошибся.** Я собрал релиз с `(.-fieldSizing style)` и посмотрел в бандл:

```js
!C.h(getComputedStyle(a).fieldSizing,"content")
```

Свойство осталось неперeименованным, предупреждений ноль. Extern inference в shadow-cljs (в релизных сборках включена по умолчанию) вывела тип и добавила объявление сама. То есть «релизная сборка молча уедет на измерение» — неправда, и комментарий к функции я поправил.

Строковую форму оставил: `getPropertyValue` — штатный способ спросить вычисленное свойство по его CSS-имени, и он не зависит от того, справилась инференция или нет. Но обосновывать её страшилкой было нечестно.

src/client/adapters/progress_store.cljs
u473t8: Agree. But what is better -- subtype as a prefix or explicit subype?

src/client/adapters/progress_store.cljs
u473t8: Да

src/client/pages/home/actions.cljs
u473t8: почему два hätte не могут иметь разный подтип?

u473t8: Фраз ещё нет, миграции не нужны

u473t8: В остальном подтип как будто выигрвывает

src/client/adapters/progress_store.cljs
u473t8: These two are the same question as «почему два hätte не могут иметь разный подтип», so let me put them together.

The id is derived from the value. So:

- **Subtype in the id** (`phrase:hätte` vs `vocab:hätte` — today's design): two `hätte` coexist, type is readable without loading the doc, range scans stay per-type. The subtype is then immutable: it is the identity, so a wrong guess in the add form can only be fixed by delete and re-add.
- **Explicit subtype field** (`vocab:hätte` + `:kind`): the value alone is the identity, so there is exactly one `hätte` — and its kind can change. Converting a phrase into a word becomes a field write instead of a rebuild, which is the thing that has no answer at all today.

You cannot have both: "two hätte with different subtypes" requires the subtype in the id, which is the prefix again, just spelled `vocab:phrase:hätte`.

Given the duplicate case is worth four values in the whole dictionary and none we could name from use, I would take the field. It buys mutability, which the prefix can never give, and pays with a possibility nobody wants.

One behaviour to decide with it: adding a phrase whose value already exists as a word merges into that document (dedup by value), rather than creating a second one. In practice a value with a space only becomes a word if the user forced word mode, so this is rare — but it should be a decision, not a surprise.

src/client/adapters/progress_store.cljs
u473t8: Завёл: #296.

src/client/pages/home/actions.cljs
u473t8: Могут — но только если подтип попадает в `_id`. Id считается из значения (ADR-0008), поэтому при подтипе-поле `hätte` даёт один и тот же ключ независимо от того, слово это или фраза: документ ровно один.

Чтобы жили два, ключ должен различаться — то есть подтип должен быть частью ключа. Это и есть префикс, просто записанный иначе: `vocab:phrase:hätte`.

Обратная сторона размена: пока подтип в ключе, он неизменяем — «сделать из фразы слово» невозможно в принципе, только удалить и завести заново, потеряв лог повторений. Как только подтип становится полем, конверсия — это запись одного поля. Подробнее в ветке про `progress_store.cljs:85`.

u473t8: Согласен, и ещё один довод в ту же сторону, который мы пока не проговаривали: подтип-поле делает возможной конверсию. Сейчас на вопрос «ошибся при добавлении, как починить» ответа нет вообще — тип в ключе. С полем это правка одного поля в диалоге редактирования, и тогда чип в форме добавления перестаёт быть аварийным выходом: ошибка перестаёт быть необратимой.

Что нужно поменять, если идём в подтип:

- `domain/phrase.cljs`: `phrase-id` уходит, остаются доменные правила (перевод одной строкой, детект режима);
- `use_cases/phrase.cljs` сливается с `vocabulary/add!` — различие сводится к подтипу и к тому, что для фразы не запрашивается пример;
- `progress_store`: `learnable-docs` и объединяющие запросы исчезают, остаётся один диапазон `vocab:`;
- `sync.cljs`: один скан конфликтов вместо двух;
- `db/pouch.cljs`: регистрация типа `"phrase"` уходит;
- `data_export.cljs`: ветка импорта `"phrase"` уходит;
- ADR-0011 (расширение контентной адресации на `phrase:`) теряет смысл — вместо него нужна короткая запись о том, что тип живёт в теле документа;
- спека, дизайн и предложение переписываются в этой части.

Это не косметика — это откат ключевого решения ветки, поэтому сам не начинаю. Скажи «делаем здесь» — и переделаю до мержа, благо фраз ещё нет и мигрировать нечего.

u473t8:
Ручной проход после переделки на подтип — нашлось одно расхождение с тем, что я тебе раньше утверждал.

**Чип односторонний.** Презентер (`pages/home/presenter.cljs:45`) отдаёт `:show-chip?` как `(or (= :phrase mode) (some? mode-override))`, поэтому в режиме слова, пока переопределения нет, чипа на экране нет вообще. Переключить фразу в слово можно, слово во фразу — нельзя.

Практически это значит: `das Fenster`, `sich freuen`, `der Tisch` — всё, что правило про артикль объявляет словом, — фразой завести невозможно. Я раньше писал, что чип закрывает промахи эвристики в обе стороны; это неверно, закрывает одну.

Проверено в браузере: набрал `das Fenster`, чипа нет, клик некуда, запись ушла словом.

Варианты, решать тебе:

1. показывать чип всегда, когда поле непустое — тогда режим виден и переключаем в обе стороны, но чип мозолит глаза на каждом слове;
2. ничего не трогать в форме и чинить вид в диалоге редактирования, когда приедет #290 — теперь это возможно, вид лежит в поле, а не в ключе.

Второй вариант мне нравится больше: он убирает чип из горячего пути ввода и делает ошибку исправимой там, где её замечают.

u473t8:
Решение владельца: не чинить. Меньше контролов — меньше поводов ломать голову; вернёмся, если невозможность переключить вид начнёт мешать на практике.

Так что сейчас: детект решает, чип переключает только фразу в слово, а обратного пути нет — и это осознанно, а не недосмотр. Возможность появилась (вид лежит в поле, а не в ключе), но выставлять её в интерфейс мы не будем без спроса пользователя.

u473t8 (commented):
LGTM

Issue #323, opened 2026-08-15
Backend tests exist but nothing runs them


u473t8:
`test/backend/{core,examples,migrations,push,reconciliation}_test.clj` exist, but nothing
runs them. The `:test` alias (`deps.edn:24`) only adds `:extra-paths ["test"]` — no runner
dependency, no `:exec-fn` — and `.github/workflows/integration.yml:52` has a client-tests job
only.

## Scope
- Add a test runner to the `:test` alias and a documented command to run it.
- Add a "Backend Tests" job to `integration.yml`.
- Fix whatever the suite turns out to be failing on after this long unrun.

## Acceptance
- One command runs the backend suite locally.
- CI fails when a backend test fails.

Issue #304, opened 2026-08-15
Strip the CouchDB proxy-auth headers in every nginx location


u473t8:
The three `proxy_set_header X-Auth-CouchDB-*  ""` directives that strip client-supplied
proxy-auth headers are declared at `server` level in
`infra/production/etc/nginx/sites-available/learning-app.conf:24`.

nginx inherits `proxy_set_header` into a `location` only when that location declares none
of its own. `location /` (:19), `/db/` (:52) and `/db/dictionary-db/` (:91) all declare
their own headers, so the stripping does not reach them.

Same shape in `infra/development/etc/nginx/sites-available/sprecha.localhost.conf:36`.

## Scope
- Move the three directives into an nginx snippet and `include` it in every `location`
  that proxies upstream, in both the production and development configs.
- Add a smoke assertion in `infra/staging/smoke.sh`: a request carrying
  `X-Auth-CouchDB-UserName` must not reach the upstream with that header set.

## Acceptance
- Every proxying location strips the three headers.
- The smoke check fails if a future edit reintroduces the gap.

PR #343, opened 2026-08-15
GH-306 Make the example endpoint ask who is calling, and how often

### LINKED ISSUE #306: Example generation endpoint needs a token and a rate limit
`GET /api/examples` (`src/backend/core.clj:491`) does not read the session cookie, unlike
`/auth/check` and `/api/sync/updates`. The handler calls the paid OpenRouter provider
(`src/backend/examples/provider.clj:34`). In nginx, `limit_req` is applied to `/db/` only.

## Scope
- Require the session token, as `authenticated-user-id` does elsewhere in the same file.
- Add a dedicated `limit_req` zone for `/api/examples`.

## Acceptance
- An unauthenticated request gets 401 and never reaches the provider.
- Burst requests from one client are throttled.

u473t8:
Closes #306.

`GET /api/examples` was the only route in the router that never read the session cookie, and the only
one that spends money: each request runs a provider completion. At the edge, the single `limit_req_zone`
covered `/db/` only, so the endpoint fell through to `location /` unthrottled.

## What changed

- **Backend** — the handler resolves the cookie through `authenticated-user-id` and answers `401`
  before parsing parameters and before reaching the provider. An anonymous caller is not even told
  which parameter it forgot.
- **Edge** — an `examples` zone of its own (`30r/m`, `burst=10 nodelay`, `429`), in both the production
  and development configs. Sharing `couchdb` would be wrong twice: it is sized for replication bursts,
  and replication would eat a paid endpoint's budget. A `map $status` puts `Retry-After` on the
  throttled answer only — `add_header ... always` stamps every response otherwise. The new location
  repeats the `X-Auth-CouchDB-*` stripping, since declaring any `proxy_set_header` stops the
  server-level inheritance (#304).
- **Client** — `fetch-one` states its credentials mode instead of leaning on the same-origin default,
  and the task runner now starts *after* the component that writes the auth cookie. They shared a
  dependency layer, so a queued example fetch could beat the cookie and come back 401 — a coin toss on
  every boot, and exactly how this change could have broken examples instead of protecting them.

## Verified

- Backend test over the real router and a real socket: no cookie → `401`, unknown token → `401`,
  missing `word` without a session → `401` (not `400`), valid session → `200`. Provider call count
  asserted: zero for the refused ones.
- **Negative control** — with the handler reverted, that same test reports `200` for anonymous requests
  and two provider calls. The test catches the defect, it does not merely pass.
- Backend suite by hand (nothing runs `test/backend/**` in CI — #323): 25 tests, 59 assertions, 0 failures.
- Throttling against a real nginx running these exact directives against a counting upstream:
  11 answered, the rest `429` with `Retry-After: 5`, upstream hit count exactly the 11. Successful
  answers carry no `Retry-After`.
- Both repo configs load in nginx (test certs and ports substituted for the production one).
- Client suite: 182 tests, 433 assertions, 0 failures.

## Not proven

- No end-to-end run in a browser against the stand: generating a real example spends provider money,
  so the client path is argued from code (single caller, relative URL, and the service worker's fetch
  handler ignores `/api/` entirely) plus the passing suite, not from a live 200.
- The throttle was proven in an isolated nginx, not on the shared dev stand, whose installed config is
  a root-owned copy. **Deploying it needs the owner**: copy `infra/development/...` into
  `/etc/nginx/sites-available/` and reload.
- `openspec` `design.md` was deliberately skipped; the reasoning is short enough to live in the proposal.

### DIFF
--- a/infra/development/etc/nginx/sites-available/sprecha.localhost.conf
+++ b/infra/development/etc/nginx/sites-available/sprecha.localhost.conf
@@ -1,3 +1,15 @@
+# Example generation spends money at the LLM provider on every request. The
+# limit lives here as well as in production so it is exercised on the stand
+# instead of first being met by users.
+limit_req_zone $binary_remote_addr zone=examples:10m rate=30r/m;
+
+# `add_header ... always` fires on every response, so the value is emptied for
+# anything but a throttled one — nginx omits a header whose value is empty.
+map $status $examples_retry_after {
+    429     5;
+    default "";
+}
+
 # The dev stand. Plain http on purpose: browsers resolve *.localhost to
 # loopback themselves and treat it as a secure context, so Secure cookies
 # and the service worker work without certificates. The Cloudflare tunnel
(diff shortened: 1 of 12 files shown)


u473t8 (commented):
LGTM

PR #344, opened 2026-08-15
GH-304 Strip the CouchDB proxy-auth headers in every nginx location

### LINKED ISSUE #304: Strip the CouchDB proxy-auth headers in every nginx location

u473t8:
Closes #304.

## Defect

The three `proxy_set_header X-Auth-CouchDB-* ""` directives sat at `server` level in both nginx
configs. nginx passes `proxy_set_header` into a `location` only when that location declares none of
its own, and every proxying location declares at least `Host` or `Upgrade` — so the clearing
reached none of them. `/db/` was never affected: it assigns all three from its `auth_request`
variables, which replaces whatever the client sent.

## Fix

- The clearing moves into `snippets/couchdb-proxy-auth-strip.conf` (shipped by the infra deb, and
  copied on the dev stand — the include is relative, so nginx resolves it next to `nginx.conf`).
- Every proxying location includes it, except `/db/`, where the `auth_request` assignment is the
  strip; including it there would send each header twice. A comment now says so.
- The `server`-level copies stay as a net for a future location that declares no headers at all.
- `infra/staging/smoke.sh` gains two assertions: one stops the app unit, listens on its port and
  reads the request nginx actually forwards; one walks `nginx -T` for a proxying location that
  neither clears nor assigns the three headers. The staging image gains `netcat-openbsd` for the
  first.

## Verification

Measured on a throwaway nginx (1.24) built from these very configs — only listen port, server_name,
TLS and upstream addresses rewritten, every header directive verbatim — proxying to an upstream
that reports the headers it received:

| location | before | after |
| --- | --- | --- |
| `/` | forwards `attacker` / `_admin` / `forged` | clean |
| `/db/dictionary-db/` | forwards them | clean |
| `/shadow-cljs/` (dev) | forwards them | clean |
| `/api/sync/updates` | forwards them | clean |
| `/db/` | `alice` / `users` / `tok-abc` | unchanged, one line each |

Both nginx configs were checked this way, before and after. The two smoke assertions were dry-run
against the same stand: both fail on the pre-fix config and pass on the fixed one, so neither is
vacuous.

Not verified here: a full `infra/staging/run.sh` container run (needs the built jar) — the smoke
check on this PR covers it. The shared dev stand was not touched; nothing was applied anywhere
outside a throwaway prefix in `/tmp`.

### DIFF
--- a/infra/staging/Dockerfile
+++ b/infra/staging/Dockerfile
@@ -15,11 +15,14 @@ RUN printf '%s\n' \
         'Acquire::https::Timeout "30";' \
     > /etc/apt/apt.conf.d/80-staging-timeouts
 
-# systemd as PID 1 plus tooling for the repo setup below.
+# systemd as PID 1 plus tooling for the repo setup below. netcat-openbsd is for
+# the smoke alone: it stands in for an upstream so an assertion can read the
+# request nginx actually forwards. It never enters the production deb.
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         systemd systemd-sysv dbus \
         curl ca-certificates gnupg apt-transport-https openssl \
+        netcat-openbsd \
     && rm -rf /var/lib/apt/lists/*
 
 # CouchDB apt repository — the same source admin-setup.sh --bootstrap adds.
--- a/docs/dev/development-setup.md
+++ b/docs/dev/development-setup.md
@@ -184,10 +184,17 @@ Restart CouchDB after copying the config.
 
 ### 4. Install the Nginx config
 
+The site config `include`s a snippet, so copy both — nginx refuses to start when
+an included file is missing. The include is relative, and nginx resolves it
+against the directory holding `nginx.conf`.
+
 ```bash
 # macOS
+mkdir -p /opt/homebrew/etc/nginx/snippets
+cp infra/development/etc/nginx/snippets/couchdb-proxy-auth-strip.conf /opt/homebrew/etc/nginx/snippets/
 cp infra/development/etc/nginx/sites-available/sprecha.localhost.conf /opt/homebrew/etc/nginx/servers/
 # Ubuntu/WSL
+sudo cp infra/development/etc/nginx/snippets/couchdb-proxy-auth-strip.conf /etc/nginx/snippets/
 sudo cp infra/development/etc/nginx/sites-available/sprecha.localhost.conf /etc/nginx/sites-available/
 sudo ln -sf /etc/nginx/sites-available/sprecha.localhost.conf /etc/nginx/sites-enabled/
 ```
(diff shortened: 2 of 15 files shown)


infra/staging/Dockerfile
```diff
@@ -15,11 +15,14 @@ RUN printf '%s\n' \
...
 
-# systemd as PID 1 plus tooling for the repo setup below.
+# systemd as PID 1 plus tooling for the repo setup below. netcat-openbsd is for
+# the smoke alone: it stands in for an upstream so an assertion can read the
+# request nginx actually forwards. It never enters the production deb.
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         systemd systemd-sysv dbus \
         curl ca-certificates gnupg apt-transport-https openssl \
+        netcat-openbsd \
```
u473t8, line 25: What task netcat-openbsd solves?

u473t8: It stands in for the upstream so the assertion can read the request nginx actually forwards.

The problem it solves: no response tells you which headers reached the upstream. CouchDB
answers a spoofed request exactly as it answers an anonymous one, so a black-box check
cannot distinguish "the header was stripped" from "the header arrived and was ignored" —
which is the whole point of GH-304.

So the assertion stops `learning-app-run`, listens on 8083 with `nc`, sends a request
carrying `X-Auth-CouchDB-UserName: attacker`, and greps the captured request. It also
asserts the capture happened at all (`grep '^GET '`), so an empty capture cannot read as
clean, and restarts the app afterwards, waiting until it answers again.

Alternatives considered: a second nginx or a tiny HTTP server in the image would need more
than netcat, not less; `nginx -T` alone proves the config's shape but not the forwarded
request, which is why both kinds of assertion are there.

It is in the staging image only — the production deb never installs it.

Issue #308, opened 2026-08-15
Unbounded queries silently truncate at 25 documents


u473t8:
Three more `find` calls run with the pouchdb-find default `limit = 25`:

- `src/client/adapters/examples.cljs:130` — `purge-by-collection!`, whose docstring promises
  "every example doc"; leaves orphans in device-db.
- `src/client/adapters/collections.cljs:6` — `list-collections` truncates at 26 collections.
- `src/client/adapters/progress_store.cljs:145` — scrubbing a deleted word-id out of
  collections misses collections past the 25th, leaving dangling ids in `:word-ids`.

## Scope
- Move all three to `db/find-all` (`lib/db/src/db.cljc:446`) or an explicit limit where a
  limit is actually intended.
- Cover the purge and the word-id scrub with tests above the boundary.

## Acceptance
- Purging a collection removes every example doc for it.
- Deleting a word removes its id from every collection.

Needs: shares the `find-all` change with the local-db migration fix.

PR #345, opened 2026-08-15
GH-307 Copy every local-db document, not the first 25

### LINKED ISSUE #307: Local-db migration copies only the first 25 documents of each type
`copy-type!` (`src/client/db_migrations.cljs:56`) calls `db/find` with no `:limit`.
pouchdb-find defaults to `limit = 25` (`node_modules/pouchdb-find/lib/index-browser.js:1348`),
so the local-db split copies at most 25 documents per type into `user-db`/`device-db`.

`run-local-db-split!` writes the migration marker regardless, so the migration cannot be
re-run: everything past the 25th word or review stays behind in the old database.

`db/find-all` already exists (`lib/db/src/db.cljc:446`).

## Scope
- Use `find-all` in `copy-type!`.
- Extend `test/client/db_migrations_test.cljs:58` — it seeds 2 docs, so the case is unseen.
  Seed more than 25 of a type and assert every doc lands in the destination.
- Decide what to do for a browser that already ran the truncated migration: the marker is
  present, the leftovers are not. A repair pass or a bumped migration id.

## Acceptance
- A local-db with 100 words migrates all 100.
- The test fails without the fix.

u473t8:
Closes #307

## Defect

`copy-type!` called `db/find` with no `:limit`; pouchdb-find defaults to 25. The local-db
split copied one page per document type and wrote its marker anyway, so everything past
the 25th word, review, example, lesson or task stayed in `local-db` and the migration
could never run again. `run-task-data-payload!` read the same way.

The test seeded 2 documents per type, well under the page, which is why nobody saw it.

## Fix

- `db/find-all` in both migration reads.
- A repair migration, `migration:local-db-split-repair`, runs both passes once more under
  its own marker, so browsers that recorded a truncated run get their remainder. The
  original marker is left alone: it still records that the split happened.
- Copying now skips ids the destination already knows, tombstones included. Without that
  guard the repair resurrects a word deleted after the truncated run — `delete-word!`
  tombstones vocab and reviews in `user-db`, and inserting a stripped document over a
  tombstone recreates it. Removing the guard fails
  `repair-does-not-resurrect-a-deleted-word` with `(not (= 2 3))`.

## Verification

Node client tests, `npx shadow-cljs compile node-test && node target/node-tests.js`.

Before the fix, with the new case seeding 30 vocab documents:

```
FAIL in (run-local-db-split-copies-past-the-default-page)
expected: (= total (count vocabs))
  actual: (not (= 30 25))
Ran 183 tests containing 434 assertions.
1 failures, 0 errors.
```

After:

```
Ran 188 tests containing 447 assertions.
0 failures, 0 errors.
```

Not proven in a real browser: verification is the node suite against PouchDB's
LevelDB adapter, not IndexedDB in Chrome.

## Notes

- #308 covers the same unbounded-`find` defect elsewhere in the codebase. This PR touches
  only `src/client/db_migrations.cljs` and its test; `lib/db` is used, not changed.
- OpenSpec change `migrate-every-local-db-document` is archived in this branch; three
  requirements are folded into `openspec/specs/data-model/spec.md`.


u473t8 (commented):
The thing is we have no users to migrate yet.

u473t8:
Тогда стоит проверить фактами, потому что вопрос не в чужих пользователях.

`run-local-db-split!` в master с 5 февраля (`a485d0b`, GH-64) — полгода. Он прогоняется на
любом устройстве, где приложение открывали после этой даты, и молча копирует по 25
документов на тип. То есть под усечение попадают не «будущие пользователи», а любая
браузерная база, где на тот момент было больше 25 слов или отзывов — включая наши
собственные телефоны и десктопы.

Данные при этом целы: `local-db` никто не удаляет (`db/destroy` зовётся только для
`user-db`, в `sync.cljs`), остаток просто лежит там, недоступный приложению.

Проверяется одной строкой в консоли на каждом устройстве, где приложением пользовались:

```js
new PouchDB('local-db').info().then(i => console.log(i.doc_count))
```

Ноль или отсутствие базы — мигрировать некого, и ремонтную миграцию я убираю: остаётся
только переход на `find-all` и тест. Ненулевое число — ремонт нужен ровно для этого случая.

Скажи, что показало, и я приведу PR к нужному варианту.

u473t8:
> The thing is we have no users to migrate yet.

Understood — dropped in f93cae8.

Removed: `run-truncated-copy-repair!` and its `migration:local-db-split-repair` marker, its entry in the migration list, its four tests (including `repair-does-not-resurrect-a-deleted-word`), and the `known-ids` guard built on `all-docs`, which had no other caller. The `run-once!` / `split-local-db!` / `rewrite-legacy-tasks!` extraction went with it — it existed only so the repair could re-run those passes — so `db_migrations.cljs` is back to its original shape and the source diff is now two lines.

What is left is the defect and its proof:

- `db/find-all` in `copy-type!` and in the task-data rewrite;
- `run-local-db-split-copies-past-the-default-page` and `run-task-data-payload-rewrites-past-the-default-page`, seeding 30 documents each.

Without the fix:

```
FAIL in (run-local-db-split-copies-past-the-default-page)
expected: (= total (count vocabs))
  actual: (not (= 30 25))

FAIL in (run-task-data-payload-rewrites-past-the-default-page)
expected: (every? :data tasks)
  actual: five tasks still carrying :word-id

Ran 184 tests containing 437 assertions.
3 failures, 0 errors.
```

With it:

```
Ran 184 tests containing 437 assertions.
0 failures, 0 errors.
```

OpenSpec follows the code: the `An incomplete migration is repaired, not skipped` and `A migration never resurrects a deleted document` requirements are gone from `openspec/specs/data-model/spec.md` and from the archived delta, leaving only `A migration reads every document it claims to migrate`. The archived proposal, design and tasks were rewritten to match; the design records the dropped repair as a decision rather than pretending it was never considered. `openspec validate --specs` is 27/27.

PR left open for you to merge.

u473t8 (commented):
LGTM

Issue #351, opened 2026-08-16
The dictionary serves only one tab at a time


u473t8:
The SQLite dictionary is reachable from one browsing context at a time. Every other
open tab of the app has no autocomplete at all, and says nothing about why.

## Mechanism

`resources/public/js/sqlite3-worker.js:24-36` takes the Web Lock
`sprecha-sqlite-opfs-sahpool` with `ifAvailable: true`, so the second context fails
fast by design. It posts `{type: "error", code: "another-tab-open"}` with a
user-ready Russian message, but `src/client/db/sqlite.cljs:57` only logs it, and
release builds have logging off. `src/client/adapters/dictionary.cljs:52` then sees
`ready? = false` and returns `[]`, so the suggestion list is simply empty.

The lock is held for the worker's whole lifetime (`sqlite3-worker.js:44-47`), so the
loser never recovers while the winner stays open.

## Observed

Five app tabs open on a phone; the OPFS database was fully present (37 MB in
`usageDetails.fileSystem`, quota 10 GB, `persisted: true`), so this is not a download
or storage failure. Suggestions returned as soon as the extra tabs were closed.

## Scope

Serve one dictionary to every context instead of one context to the dictionary:
move the SQLite owner into a `SharedWorker` and have each tab talk to it over a port.
Keep the OPFS SAH pool single-owner inside that shared worker.

Fallback matters: `SharedWorker` is unavailable in some contexts, so a dedicated
worker must still work when no shared worker can be created.

## Acceptance

- Two or more tabs of the app open at once, each with working autocomplete.
- Closing any one tab leaves the others working.
- No regression in cold-start time for a single tab.

Related: #312 covers the reporting side — a context that still cannot reach the
dictionary must say so instead of showing an empty list. Neither blocks the other.

Issue #346, opened 2026-08-15
The worktree rule cannot be followed by an isolated subagent


u473t8:
The repo rule (AGENTS.md, "Branches and Worktrees"; `.claude/rules/repo-delivery.md`) is
`gh issue develop <n>` → `git worktree add .claude/worktrees/<slug> <branch>` → `EnterWorktree`.
A subagent launched with worktree isolation cannot follow it.

Observed three times today, filing the first wave of backlog work (#304, #306, #307):

- `git worktree add .claude/worktrees/304-...` succeeds — creation is not the problem.
- After `EnterWorktree` into it, file edits still work, but **every** Bash call is refused,
  including `git -C <path>` and even `pwd`: the guard treats only the launch worktree as
  in-bounds and resolves anything else as outside.
- `ExitWorktree` cannot undo it either — it is not available to a subagent whose cwd was
  pinned at launch.

That leaves two ways out, both wrong: switch branches inside the launch worktree, which
belongs to another task, or nest the new worktree under the launch one. The agents picked
nesting, so three branches now live at `.claude/worktrees/331-issue-flow-cost/wt-304`,
`.../wt-306` (the launch worktree itself, switched) and `.../wt-307` — linked worktrees of
the right gh-created branches, in the wrong place.

## Scope

- Decide what a worktree-isolated subagent should do, and say it in the rule: work in the
  worktree it was given, or be launched without isolation when it needs its own.
- Whatever is decided, the rule text must describe a sequence that actually runs end to end
  in that context — the current one cannot.
- Clean-up guidance should cover nested worktrees, since some already exist.

## Acceptance

- An agent following the rule as written, under worktree isolation, reaches a pushed branch
  without hitting a refused Bash call.
- No leftover worktree nested under another worktree.

u473t8:
Mechanics, so whoever takes this does not rediscover them (found while delivering #304).

**The guard is not ours.** It is not `.claude/settings.json` — the only Bash `PreToolUse` hook there
is `delivery-guard.sh`, which refuses raw `gh issue create`. It is not in the user-level settings
either. The string appears nowhere in the repository. It is built into Claude Code's worktree
isolation, and per its documentation the check cannot be turned off: no settings key, env var or
permission entry relaxes it.

**What it refuses, precisely.** Four checks: file edits, the command's resolved working directory,
git redirects (`git -C <path>` is blocked even when the path is a linked worktree of the same repo),
[...]

PR #353, opened 2026-08-16
GH-352 route workspace by dependency, not by topic

### LINKED ISSUE #352: The workspace rule routes by topic, not by dependency
AGENTS.md picks the workspace by naming topics — "sync, dictionary, migrations, schema,
anything talking to CouchDB or nginx" go to the main checkout. Topics are easy to match and
easy to over-match: a task only has to mention the dictionary to be routed to the main
checkout, whether or not it needs the stand at all.

That misrouting just happened. Work on #351 (multi-tab access to the SQLite dictionary) was
sent to the main checkout because it is dictionary work. It needed no stand: proving that two
tabs both get suggestions needs one origin, OPFS and a Web Lock, and the fixture dictionary
the browser tests already use serves that fine on `localhost`. The real database matters only
for cold-start timing.

The cost was not only a wrong directory. A background session cannot write to the shared
checkout at all — the harness blocks it and points at `isolation: "worktree"` — so the agent
stopped on its first write with nothing done.

## Scope

- State the workspace rule as a test on dependencies, not a list of topics: name the concrete
  thing that needs the stand (a live CouchDB, nginx routing, a migration against real data),
  or the work goes to a worktree.
- Say plainly what a worktree can still verify: browser behaviour on fixture data, since
  `localhost` is a secure context and OPFS, Web Locks and workers all function there.
- Record that a background coordinating session has no third option. "The coordinator stays
  in the main checkout and delegates" plus "background writes to the shared checkout are
  blocked" means every delegated edit must go to a worktree.
- Keep `.claude/rules/repo-delivery.md` consistent with AGENTS.md; it restates the same rule.

## Acceptance

- The rule can be applied without knowing which subsystem a task touches.
- A reader can tell, from the rule alone, whether a two-tab browser check belongs in a
  worktree.
- The background-session constraint is written down where the delegation rule lives, not
  discovered when a write is refused.

Related: #346 recorded the previous failure of the same rule, from the opposite direction —
an agent that could not reach a worktree at all.

u473t8:
Closes #352.

## Problem

The workspace rule picked the directory by naming topics — "sync, dictionary, migrations, schema, anything talking to CouchDB or nginx" went to the main checkout. A topic list over-matches: #351 (multi-tab access to the SQLite dictionary) was routed to the main checkout for being dictionary work and depended on nothing shared. Two tabs sharing an OPFS database through a Web Lock needs one origin and the fixture dictionary the browser suite already serves; the real database matters only for cold-start timing.

The cost was not just a wrong directory. A background session cannot write to the shared checkout — the harness blocks it and points at worktree isolation — so the delegated agent stopped on its first write having done nothing.

## Change

- The rule is a dependency test: name the shared thing the work needs (a live CouchDB, nginx routing, a migration over real data) or the work goes to a worktree. Subsystems stay as examples, never as the test.
- States what a worktree still proves: browser behaviour on fixture data, since it serves on `localhost` — a secure context — so OPFS, Web Locks, service workers and workers all run.
- Records next to the delegation rule that a background coordinator has no third option, so it is not first learned from a refusal. Launch delegated edits with `isolation: "worktree"`; entering a worktree as coordinator is not the way out, since the pin is inherited (#346).
- `.claude/rules/repo-delivery.md` restates the same rule and is kept in step with AGENTS.md.

## OpenSpec

Applicable and used — the repo tracks its own process rules as specs, and the topic list lived in one. Change `route-workspace-by-dependency` modified two requirements and is archived in this branch:

- `repo-agent-infrastructure` — "Worktrees live where the agent tool puts them": the full-stand scenario stops enumerating subsystems, plus scenarios for an over-matched subsystem and for a fixture-data browser check.
- `repo-delivery-workflow` — "A branch for tracked work is linked to its issue": the background-coordinator constraint.

`design.md` was skipped deliberately; the schema requires only tasks, and a rule rewording has no design to record.

## Verification

Docs only — no application code, build or runtime change. Verified by reading, not by tests:

- Both files re-read whole. One contradiction was found and fixed while writing: the first draft offered "enter a worktree before launching" as an alternative, which contradicts the coordinator rule and the inherited-pin warning (#346). It now names that as the way out not to take.
- New text does not conflict with delegation, agent pinning, the `gh issue develop` + `git worktree add` order, or the do-not-route-around-a-refusal rule.
- Every file, script and hook the touched text names exists; the fixture dictionary claim was checked against `test/fixtures/dictionary/lemmas.edn`, which the browser suite serves.
- `openspec validate --specs --strict`: 27 passed, 0 failed.

Not verified: nothing was run, because nothing here executes.


u473t8:
LGTM

Issue #312, opened 2026-08-15
Dictionary worker failures never reach the user


u473t8:
The dictionary worker already produces a user-ready Russian message —
"Приложение уже открыто в другой вкладке" (`resources/public/js/sqlite3-worker.js:30`) —
but the client only logs it (`src/client/db/sqlite.cljs:57`), and in a release build logging
is off entirely.

Meanwhile `src/client/adapters/dictionary.cljs:53` returns `[]` when the worker is not ready,
so autocomplete is silently empty with no explanation and no retry once the worker comes up.

## Scope
- Carry dictionary state (loading / ready / failed, with the worker's message) into app state.
- Show a banner for the failed case; show that the dictionary is still loading rather than
  "no matches"; retry the pending query once ready.

## Acceptance
- Opening a second tab explains what happened instead of an empty suggestion list.
- Typing before the worker is ready eventually shows suggestions without re-typing.

Needs: release logging (error visibility) makes this verifiable in a real build.

Issue #320, opened 2026-08-15
A crashed dictionary worker leaves callers waiting forever


u473t8:
`src/client/db/sqlite.cljs:7` attaches a fresh `message` listener per `exec` call and removes
it only when a matching id arrives. When the worker crashes (`error` handler, :68) the pending
promises are never rejected and their listeners are never removed — the caller waits forever
and listeners accumulate.

## Scope
- Route messages through one dispatcher keyed by id.
- Reject all pending calls on `error`, and add a timeout.

## Acceptance
- Killing the worker rejects in-flight calls instead of hanging.
- Listener count does not grow with query count.

Issue #311, opened 2026-08-15
Release builds record no errors at all


u473t8:
`src/client/logging.cljs:16` sets `:glogi/root :off` in the release build. Every error path
in the client ends in `log/error` or `log/warn` — boot failure, sync failure, add-word
failure, dictionary worker crash — so in a release build none of them are recorded anywhere,
and none of them surface in the UI.

The result: a user hitting any of these sees nothing, and there is no artifact to look at
afterwards.

## Scope
- Decide the release log level (`:warn` looks right) and what stays off.
- Give errors a path that outlives the console: at minimum an error channel into app state
  so the UI can react; consider what #301 (device metrics) would want.
- Check against #298 (release builds still ship development instrumentation) so the two
  changes do not fight.

## Acceptance
- A forced failure in a release build leaves a trace a developer can find.

Issue #298, opened 2026-08-15
Release builds still ship the development instrumentation


u473t8:
## What

`openspec/specs/repo-browser-debugging/spec.md` states "Release builds SHALL NOT contain the instrumentation", and `src/client/instrumentation.cljs` opens by saying "a release build eliminates the whole namespace". Neither is true.

A release build (`npx shadow-cljs release app`) was inspected directly. `resources/public/js/app/main.js` contains `__metricsReset=function(` and the string `hadRecentInput`, so the observers and the page-facing API both survive. This was confirmed against the `master` versions of the files, not only against work in progress, so it is not a regression from #297.

`goog.DEBUG` itself is substituted correctly — the identifier appears zero times in the release output — so the constant folds, yet dead-code elimination still keeps the namespace. Writing the DEBUG guard as `goog/DEBUG` rather than `js/goog.DEBUG` made no difference.

## Why it matters

Dead weight in the production bundle is the small part. The larger part: two documents assert a property the build does not have, and #297 had to relax its own spec wording because of it. A claim about release contents is only worth what an inspection of the output says.

## Notes

`:release-stubs` (already used for `debux.core`) is the obvious first thing to try and does not work as-is: adding `instrumentation` to the set fails the build with "The required namespace \"instrumentation\" is not available", because the generated stub carries no vars for the callers to reference.

#297 avoids the related problem for the measurement library by fetching it at runtime instead of importing it, and verifies that by grepping the bundle. The same trick does not apply to the namespace itself, which is called directly from the render path.

Needs: #297

PR #354, opened 2026-08-16
GH-351 elect one dictionary holder instead of one per tab

### LINKED ISSUE #351: The dictionary serves only one tab at a time

u473t8:
Closes #351.

## Defect

`sqlite3-worker.js` took the `sprecha-sqlite-opfs-sahpool` Web Lock with
`ifAvailable: true` and held it for the worker's life. One tab won; every other
tab was answered `{type: "error", code: "another-tab-open"}`, which the client
only logged — and release builds log nothing — so `completions` returned `[]`
and the suggestion list was silently empty.

## Not a SharedWorker

The issue proposed moving the owner into a `SharedWorker`. That was built first
and measured on Chrome 144:

| API inside `SharedWorkerGlobalScope` | value |
|---|---|
| `navigator.storage.getDirectory` | `function` |
| `FileSystemFileHandle` | `function` |
| `FileSystemFileHandle.prototype.createSyncAccessHandle` | `undefined` |
| `Worker` | `undefined` |
| `navigator.locks` | `object` |

`installOpfsSAHPoolVfs` fails there with `Missing required OPFS APIs`, and the
obvious repair — a shared worker spawning a dedicated worker to hold the pool —
is closed off by `Worker` being undefined. The pool is dedicated-worker
territory, one holder at a time. A backgrounded tab cannot hold it either:
Android freezes one whole about a minute after it leaves the screen — two tabs,
one query every 10 s, database behind the backgrounded one: 12 ms, 14, 17, 22,
31, 31, then nothing from 60 s on, 14 consecutive timeouts, no error and no
rejected promise.

## What this does instead

The dictionary belongs to the tab being typed into. A tab takes the
`sqlite-opfs-sahpool` lock (no `ifAvailable`) when it comes to the foreground,
opens the database, serves only itself, and closes and releases when it leaves.
One stretch of holding the lock is a *turn*; outside its turns a tab has no
database.

- Foreground is `visibilityState === "visible" && hasFocus()`, computed by the
  page and reported to the worker as one boolean. Visibility alone cannot pick
  between two tabs visible side by side.
- No tab-to-tab traffic at all: no shared owner, no channel, no query crossing
  a tab boundary. That is what makes the freeze irrelevant — the tab holding
  the database is one the system is not about to freeze.
- **A query asked without a turn is answered, not queued.** It comes back with
  an empty result — the same shape a real empty answer has — and is not
  replayed when the turn arrives; the next keystroke is what asks again. An
  open failure is still returned as an error, so a broken dictionary stays
  distinguishable from a word the dictionary does not have.
  The windows this gives up are ~14 ms to take the lock back and ~94 ms for a
  tab's first turn, against a 100 ms trailing suggest debounce and a keystroke
  every 120-180 ms, so a warm start has its turn before the query is sent. A
  cold start is the visible case: the dictionary is still downloading and those
  keystrokes get an empty list. Accepted for now rather than kept behind a
  queueing mechanism.
- `:dictionary/ready?` stays on the port as a reading, not a gate — an empty
  answer alone cannot say whether the prefix has no completions or this tab has
  no dictionary. #312 is what has to describe that to the user.
- Between turns the pool stays installed and paused: `unpauseVfs` 3-5 ms plus
  open 1-2 ms, against `installOpfsSAHPoolVfs` 13-27 ms plus a manifest fetch.
  The engine loads on a tab's first turn, not at boot.
- Gone: the `another-tab-open` error and its user-facing message. There is no
  losing tab left to describe.

ADR-0012 supersedes ADR-0002 in part (WASM cold start is no longer off the
critical path). OpenSpec change `dictionary-follows-focus` is archived in this
branch.

## Verified

- `test/browser/dictionary-focus.spec.js`, eight specs: the foreground tab has
  it and a background one does not; two visible tabs, the one with the keyboard
  gets it; a query asked while waiting comes back empty and is not replayed;
  the tab that leaves gives it back; focus flickering leaves the foreground tab
  working; no tab in front is a resting state and the first one back takes it;
  a tab that boots in the background has nothing until it is looked at; closing
  the holder releases it.
- The "not replayed" spec **fails on the queueing worker** — at the assertion
  that the arriving turn delivers no old answer — and passes here.
- Full browser suite 26/26 (desktop + mobile), node tests 191/191, including
  unit tests for the page protocol in `test/client/db/sqlite_test.cljs`: the
  query goes out with no readiness message, no rows is an ordinary answer, an
  error rejects, and `ready?` follows the turns.

## Not proven here

- Cold start on the full dictionary, and behaviour on a phone. The suite runs
  the ~28 KB fixture in headless Chrome on localhost, so the one window where
  an empty answer is visible to a user is not measured by any test.
- Playwright reproduces neither half of the foreground: every page reports
  `visible` and `hasFocus() === true`, `bringToFront()` changes neither, and
  `Emulation.setPageVisibilityOverride` is gone. Both halves are shimmed at the
  document and the real events dispatched, which drives the whole app path —
  but the Android freeze and real OS focus are covered by no test.
- Offline: the suite asserts nothing about the service worker.
  `/js/sqlite3-dictionary.js` is precached, but whether the SW intercepts a
  worker's `importScripts` was not measured — same exposure as the existing
  `importScripts("sqlite3.js")`.

### DIFF
--- a/resources/public/js/sqlite3-dictionary.js
+++ b/resources/public/js/sqlite3-dictionary.js
@@ -11,15 +11,12 @@
 // turns it has no database at all.
 //
 // Coming to the foreground and having the database are not the same instant,
-// and queries asked in the gap wait for it. A tab's first turn has to load the
-// engine and install the pool (~94 ms); a later one waits for the tab in front
-// to hand the lock back (~14 ms). The user is typing through all of it — this
-// is the tab they are looking at — so what waits is mostly the *active* tab's
-// own keystrokes, not some background tab hoarding work. A tab that has just
-// left the foreground can add a straggler or two, where a query was in flight
-// or the suggest debounce ticked over after it left. Either way the request
-// waits for the turn and is never dropped — see `request` for why that matters
-// more than it looks.
+// and a query asked in the gap is answered with what the tab has at that
+// instant: nothing. Nothing is queued and nothing is replayed — see `request`.
+// A tab's first turn has to load the engine and install the pool (~94 ms); a
+// later one waits for the tab in front to hand the lock back (~14 ms). Both
+// are shorter than the 100 ms debounce in front of the query, so a warm start
+// has almost always finished before the first query arrives.
 //
 // On screen alone is not enough. Two tabs can be visible at the same time —
 // split screen, two windows side by side — and only one of them is being
@@ -79,16 +76,9 @@ let yieldTurn = null;
 
 // Where this worker is, as the message its page is sent. `loading` covers both
 // "starting up" and "waiting for another tab to finish", because from the
-// page's side they are the same thing: no dictionary yet, ask anyway.
+// page's side they are the same thing: no dictionary here, so no completions
+// from here either.
 let lifecycle = { type: "loading" };
-
-// One promise that every request waiting for a turn awaits, and the resolve
-// that fires it. Shared on purpose: a second request arriving while the first
-// is still waiting joins the same promise instead of adding another, so
-// `announceSettled` always belongs to the `settling` currently in hand and
-// there is never a second one to overwrite it.
-let settling = null;
-let announceSettled = null;
 
 let broadcast = () => {};
 
@@ -186,7 +176,7 @@ async function close() {
 }
 
 
-function answer(msg) {
+function answer(db, msg) {
   try {
     return { id: msg.id, result: db.exec(msg) };
   } catch (err) {
@@ -195,44 +185,42 @@ function answer(msg) {
 }
 
 
-// What to say to one request, or null if this tab cannot say anything yet.
-// Only the decision — who receives it, and when, is the caller's business.
-function respond(msg) {
-  if (db) return answer(msg);
+// What to say to one request, given what this tab has. An empty result is the
+// same shape a real empty answer has — `resultRows` is an array — so the page
+// reads it as a prefix with no completions. An error is not flattened into
+// one: a dictionary this tab could not open has to stay distinguishable from
+// a word it does not contain (#312).
+function respond(db, lifecycle, msg) {
+  if (db) return answer(db, msg);
   if (lifecycle.type === "error") return { id: msg.id, error: lifecycle.message };
-  return null;
+  return { id: msg.id, result: [] };
 }
 
 
-// What everything waiting for a turn blocks on, created on first demand.
-function whenSettled() {
-  if (!settling) settling = new Promise((resolve) => { announceSettled = resolve; });
-  return settling;
+// The response to one request, decided from what this tab has at that instant.
+// Nothing is queued and nothing is replayed: without a turn the answer is an
+// empty list, and being asked again takes another keystroke.
+//
+// What that loses is bounded by the debounce in front of it. Taking the lock
+// back from the tab in front costs ~14 ms and a first turn ~94 ms, against a
+// 100 ms trailing debounce and a keystroke every 120-180 ms, so on a warm
+// start the turn has begun before the query is sent. A cold start is the one
+// window a user actually sees: the dictionary is still downloading, and every
+// keystroke until it lands gets an empty list. Holding those queries buys
+// that one case and costs a mechanism to do it; the empty answer is accepted
+// for now, and #312 is where "the dictionary is not here" stops looking like
+// "no such word".
+function request(msg) {
+  return respond(db, lifecycle, msg);
 }
 
 
-// The response to one request: given straight back if this tab has a turn,
-// otherwise awaited until it gets one and asked again. Never dropped —
-// answering early with an empty list looks exactly like a word the dictionary
-// does not have.
-async function request(msg) {
-  const now = respond(msg);
-  if (now) return now;
-  await whenSettled();
-  return respond(msg);
-}
-
-
-// Call only with a decided state — `ready` or `error`. Everything waiting
-// wakes and asks `respond` again, so a `loading` passed here would hand those
-// requests an answer that does not exist yet.
-function settle(state) {
+// Where this worker is, told to the page. It feeds the page's log and its
+// instrumentation and nothing here waits on it, so `loading` is as reportable
+// as `ready` and `error`.
+function report(state) {
   lifecycle = state;
   broadcast(state);
-  const wake = announceSettled;
-  announceSettled = null;
-  settling = null;
-  if (wake) wake();
 }
 
 
@@ -258,23 +246,22 @@ function acquire() {
       } catch (ignored) { /* nothing better to try */ }
       // Released by returning: this tab could not open the database, and the
       // next one — online where this one was not, say — deserves its turn.
-      settle({ type: "error", message: String(err) });
+      report({ type: "error", message: String(err) });
       return;
     }
     // Left the foreground while the database was opening. Nothing has been
     // promised to the page yet, so it just goes straight back.
     if (foreground) {
-      settle({ type: "ready" });
+      report({ type: "ready" });
       // Assigned with no await in between, so no report from the page can
       // slip past between the check above and this handle existing.
       await new Promise((done) => { yieldTurn = done; });
       yieldTurn = null;
     }
     await close();
-    // Back to waiting: queries asked from here on are held, not answered from
-    // a database this tab no longer has.
-    lifecycle = { type: "loading" };
-    broadcast(lifecycle);
+    // Back to waiting: queries asked from here on are answered empty, not
+    // from a database this tab no longer has.
+    report({ type: "loading" });
     // In front again already — the tab came back while this turn was being
     // wound up. The request queues behind this callback and is granted as
     // soon as it returns, so nothing is lost to the gap.
--- a/resources/public/js/sqlite3-dictionary.js
+++ b/resources/public/js/sqlite3-dictionary.js
@@ -1,0 +1,315 @@
+// The dictionary belongs to the tab being typed into.
+//
+// `opfs-sahpool` takes exclusive sync access handles over its pool files, so
+// one browsing context at a time can have the database open. Which one should
+// it be? The foreground tab: on screen and holding the keyboard. A tab takes
+// the `sqlite-opfs-sahpool` lock when it comes to the foreground and gives it
+// back when it leaves.
+//
+// A *turn* is one stretch of holding that lock, from the grant to the release.
+// Everything this file does with the database happens inside a turn; between
+// turns it has no database at all.
+//
+// Coming to the foreground and having the database are not the same instant,
+// and queries asked in the gap wait for it. A tab's first turn has to load the
+// engine and install the pool (~94 ms); a later one waits for the tab in front
+// to hand the lock back (~14 ms). The user is typing through all of it — this
+// is the tab they are looking at — so what waits is mostly the *active* tab's
+// own keystrokes, not some background tab hoarding work. A tab that has just
+// left the foreground can add a straggler or two, where a query was in flight
+// or the suggest debounce ticked over after it left. Either way the request
+// waits for the turn and is never dropped — see `request` for why that matters
+// more than it looks.
+//
+// On screen alone is not enough. Two tabs can be visible at the same time —
+// split screen, two windows side by side — and only one of them is being
+// typed into; `document.hidden` is false for both, so it would hand the
+// database to whichever tab happened to take the lock first rather than the
+// one the user is working in. Focus is what distinguishes them, so the page
+// reports `visibilityState === "visible" && hasFocus()`, and this file is
+// told the answer as one boolean without being told how it was reached.
+//
+// The visibility half is not a preference, it is what Android leaves us.
+// Measured with two tabs and one query every 10 s while the tab holding the
+// database sat in the background: 12 ms, 14, 17, 22, 31, 31 — and from 60 s
+// on, nothing at all, 14 consecutive timeouts until it was brought back to
+// the front. A backgrounded tab is frozen whole, worker and message delivery
+// together, so anything that leaves the database behind a backgrounded tab
+// has built a dictionary that stops answering after a minute.
+//
+// `visibilitychange` and `blur` both fire the moment the tab stops being the
+// one in front, and the freeze lands 50-60 s later, so the handover has three
+// orders of magnitude more time than it needs. If a tab is killed outright
+// instead, the browser releases its lock and the next tab takes over — the
+// same path as closing one.
+//
+// When no tab is in the foreground — the whole browser window is behind
+// something else — no tab holds the database. There is nobody to ask, and the
+// first tab to come back takes it in ~14 ms.
+
+const SAH_POOL_LOCK = "sqlite-opfs-sahpool";
+
+const workerParams = new URL(self.location.href).searchParams;
+const telemetry = workerParams.has("telemetry");
+
+// The engine, loaded once for the worker's life. The pool and the open
+// database come and go with the lock; this does not.
+let sqlite3 = null;
+let engine = null;
+
+// Installed on this tab's first turn and kept, paused, between turns:
+// `unpauseVfs` costs a third of a fresh install and asks the disk for nothing
+// the pool already knows.
+let pool = null;
+let poolFile = null;
+
+// Set only while this tab holds the lock. Its presence is what "this tab has
+// the dictionary" means everywhere below.
+let db = null;
+
+// What the page last said about itself: on screen and holding the keyboard.
+// Nothing is taken before it says anything, so a tab that boots in the
+// background never takes a database it would be frozen holding.
+let foreground = false;
+
+// The lock: `queued` while waiting for a turn, `yieldTurn` set while holding
+// one — calling it ends the turn, which is how the lock is given back.
+let queued = false;
+let yieldTurn = null;
+
+// Where this worker is, as the message its page is sent. `loading` covers both
+// "starting up" and "waiting for another tab to finish", because from the
+// page's side they are the same thing: no dictionary yet, ask anyway.
+let lifecycle = { type: "loading" };
+
+// One promise that every request waiting for a turn awaits, and the resolve
+// that fires it. Shared on purpose: a second request arriving while the first
+// is still waiting joins the same promise instead of adding another, so
+// `announceSettled` always belongs to the `settling` currently in hand and
+// there is never a second one to overwrite it.
+let settling = null;
+let announceSettled = null;
+
+let broadcast = () => {};
+
+
+// Every phase the page's instrumentation reports, startup and per-turn alike.
+// Nothing is accumulated here — the page keeps the list, this only reports.
+async function measure(phase, fn, extras) {
+  if (!telemetry) return fn();
+  const t0 = performance.now();
+  try {
+    const result = await fn();
+    const msg = { type: "phase", phase, status: "ok", durationMs: Math.round(performance.now() - t0) };
+    if (extras) Object.assign(msg, extras(result));
+    broadcast(msg);
+    return result;
+  } catch (err) {
+    broadcast({ type: "phase", phase, status: "error",
+                durationMs: Math.round(performance.now() - t0), reason: String(err) });
+    throw err;
+  }
+}
+
+
+// Loaded on this tab's first turn, not at startup: a tab that never comes to
+// the front never pays for the engine, and a phone with several tabs open is
+// exactly where that matters. Kept afterwards, so later turns cost only the
+// pool and the database.
+function loadEngine() {
+  if (engine) return engine;
+  engine = (async () => {
+    const dir = workerParams.get("sqlite3.dir");
+    importScripts(dir ? `${dir}/sqlite3.js` : "sqlite3.js");
+    sqlite3 = await measure("wasm-init", () => sqlite3InitModule());
+  })();
+  return engine;
+}
+
+
+async function installPool() {
+  pool = await measure("pool-install", () =>
+    sqlite3.installOpfsSAHPoolVfs({ name: "opfs-sahpool", initialCapacity: 3 }));
+
+  const manifest = await measure("manifest", () =>
+    fetch("/dictionary/manifest").then(r => r.json()));
+  const hash12 = manifest.hash.slice(0, 12);
+  poolFile = `/dict.${hash12}.sqlite`;
+
+  // Unlink stale versions first so their handles become available for import.
+  for (const name of pool.getFileNames()) {
+    if (name !== poolFile && name.startsWith("/dict.") && name.endsWith(".sqlite"))
+      pool.unlink(name);
+  }
+
+  if (!pool.getFileNames().includes(poolFile)) {
+    const buffer = await measure("download",
+      () => fetch(`/dictionary/${manifest.filename}`).then(r => r.arrayBuffer()),
+      ab => ({ bytes: ab.byteLength }));
+    await measure("import", () => pool.importDb(poolFile, new Uint8Array(buffer)));
+  } else if (telemetry) {
+    broadcast({ type: "phase", phase: "cache-hit", status: "ok", durationMs: 0, hash12 });
+  }
+
+  await measure("cleanup", () => pool.reduceCapacity(1));
+}
+
+
+async function open() {
+  await loadEngine();
+  if (pool) await measure("unpause", () => pool.unpauseVfs());
+  else await measure("init", installPool);
+  db = await measure("db-open", () =>
+    new sqlite3.oo1.DB({ filename: poolFile, vfs: "opfs-sahpool" }));
+}
+
+
+// Hands the file handles back so the next tab can take them. `pauseVfs` throws
+// while any connection is open, so the close is not optional.
+//
+// `db` is dropped before anything that can throw, and the throw is swallowed,
+// because `db` is what tells the rest of this file that the lock is ours. If
+// `close()` rejected, the lock callback would reject with it: the lock would
+// go to the next tab while `request()` kept calling `exec` on a pool this tab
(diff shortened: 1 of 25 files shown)


u473t8:
Independent re-measurement of the claim that killed the `SharedWorker` design, on Chrome 151
for Android (the PR's own probe ran on Chrome 144, so this is a second version as well as a
second device):

| | SharedWorker | Dedicated worker |
|---|---|---|
| `navigator.storage.getDirectory` | function | function |
| `FileSystemFileHandle` | function | function |
| `FileSystemFileHandle.prototype.createSyncAccessHandle` | **undefined** | function |
| `Worker` | **undefined** | function |
| `navigator.locks` | object | object |
| `BroadcastChannel` | function | function |

Both blocked paths reproduce: no synchronous file access inside a shared worker, and no way
to delegate to a dedicated one from there. The elected-holder design stands on measurement,
not on preference.

u473t8:
Three ways to let several tabs read one dictionary were measured before settling on the
elected holder. Chrome 151, the real 36 966 400-byte dictionary, `completions-sql` with real
data (158 294 lemmas), median of 21 runs. The vendored build is sqlite 3.53.1, byte-identical
to the upstream release, and it already ships every VFS discussed here.

**1. The `opfs` / `opfs-wl` VFS — supported upstream for exactly this, but not reachable here.**
Both require `SharedArrayBuffer` and `Atomics`, so both need cross-origin isolation; the app
currently runs with `crossOriginIsolated: false`, and the installer fails silently in that
state. They also need `sqlite3-opfs-async-proxy.js`, which is not vendored. Cost if enabled:
`COOP: same-origin` + `COEP: require-corp` across the site, meaning every third-party asset
must serve CORP/CORS. Performance cost measured on a cross-origin-isolated stand: a fixed
~11 ms per statement that touches the file, from the async-proxy `xLock`/`xUnlock` round trip.
Ten statements cost 112.8 ms loose but 11.7 ms inside one transaction. Autocomplete runs one
`exec` per keystroke, so prefix "fen" goes 2.1 → 13.6 ms and "f" 59.8 → 74.4 ms.

**2. Read-only handles per tab — full speed, but a fork and three measured risks.**
`createSyncAccessHandle({mode: "read-only"})` genuinely allows several simultaneous readers,
and two tabs, each with its own sqlite instance, both answered real queries at full speed
(+11 % on heavy prefixes, none on light ones). It needs a patched `OpfsSAHPool`: the blocker
is `sqlite3.js:13709`, where `getAssociatedPath` calls `sah.truncate()` on a free pool slot.
Measured risks: a failed install calls `removeVfs()`, which deletes the VFS root — one failure
costs a re-download of the whole dictionary; `acquireAccessHandles` has a handle-leak race that
reproduced non-deterministically; and any live reader blocks every writer, so updating the
dictionary needs all tabs to let go. Roughly 50-80 lines of patch inside a vendored file that
is currently pristine.

**3. Elected holder (this PR).** BroadcastChannel round trip is 0.8-0.9 ms; a fully proxied
query costs 1-3 ms over a direct one. Memory is ~0.5 MB per extra tab against ~12-18 MB for a
tab running its own instance. The real cost is serialisation: two tabs querying at the same
instant take 2x, since one holder answers both.

Coordination does not disappear under any option: even independent readers need it to
bootstrap an empty pool and to replace the dictionary, because a reader blocks the writer.
Given that two tabs typing simultaneously is not a real usage pattern while five idle tabs
very much is, paying 1-3 ms and 0.5 MB beats paying 11 ms per keystroke or forking the vendored
engine.

Not measured: browsers other than Chrome 151, live dictionary replacement with several tabs
open, and battery or low-end hardware.

adr/0012-dictionary-holder-elected-by-web-lock.md
```diff
@@ -0,0 +1,59 @@
...
+what runs.
+
+## Decision
+
+Every browsing context keeps a dedicated worker of its own, and the contexts elect which
+worker holds the database. The election is the pool lock itself, requested without
+`ifAvailable`: the worker that takes `sprecha-sqlite-opfs-sahpool` opens the file and serves
+every context over a `BroadcastChannel`; the rest wait on the same lock and relay their
+queries meanwhile, without initialising SQLite WASM or the pool. When the holder's context
+goes away the browser releases the lock, the next waiter opens the same file and announces
```
u473t8: And when holder context goes away?
What will happen to queued queries?

openspec/changes/archive/2026-08-16-one-dictionary-for-every-tab/design.md
```diff
@@ -0,0 +1,133 @@
...
+| `Worker`                                                     | `undefined` |
+| `navigator.locks`                                            | `object`    |
+
+`installOpfsSAHPoolVfs` fails there with `Missing required OPFS APIs`, and the obvious
+repair — a shared worker that spawns a dedicated worker to hold the pool for it — is closed
+off by `Worker` being undefined: nested workers are not available in a shared worker in
+Chrome. A `ServiceWorker` is worse still: it is killed and restarted at the browser's
+discretion, which would drop the pool mid-session.
+
+So the pool stays where it must be, in a dedicated worker, and the tabs elect which one.
```
u473t8: I would expect that active tab should win, is it so?

openspec/changes/archive/2026-08-16-one-dictionary-for-every-tab/proposal.md
```diff
@@ -0,0 +1,66 @@
...
+`ready? = false` and returns `[]`, so the user gets a silently empty suggestion list.
+
+Observed on a phone with five tabs open: the OPFS database was fully present (37 MB,
+`persisted: true`), so this is neither a download nor a storage failure. Suggestions
+returned as soon as the extra tabs were closed.
+
+## What Changes
+
+- The tabs elect a holder instead of the first one taking everything. The pool lock is
+  requested without `ifAvailable`, so the winner opens the database and the rest queue on
```
u473t8: Explain how is the 'ifAvailable' relevant here?

u473t8:
Reproduced against this branch (`5c01e75`, `sqlite3-dictionary.js` byte-identical to the
committed file), Chrome 151 on Android, two tabs of one origin, the real 36 966 400-byte
dictionary. Roles were confirmed from telemetry, not assumed: tab A ran the whole cold start
and announced `serving`; tab B received all eight phases at once and every query of its own
appeared on the channel as `type:"query"` with an answer from A.

With A backgrounded and B focused, querying "fen" every 10 s:

| seconds backgrounded | response |
|---|---|
| 0 | 12 ms |
| 10 | 14 ms |
| 20 | 17 ms |
| 30 | 22 ms |
| 40 | 31 ms |
| 50 | 31 ms |
| **60** | **timeout** |
| 78 … 294 | timeout, 14 consecutive |

Twenty queries sent, six answered. The cutoff is sharp, between 50 s and 60 s, preceded by a
visible slowdown from 12 to 31 ms.

The failure is silent. No `error`, no `worker-error`, no lifecycle change across five minutes
— the tab stays `ready`. `dictionary.request` never rejects, so nothing times out on its own
and the promise waits forever; the 8 s timeout above belongs to the probe, not to the app.
For a user this is "suggestions stopped appearing", with no spinner and no message. (The
never-rejecting promise is #320.)

It is not the worker alone that freezes: the holder page did not observe the querying tab's
message either, and messages queued up were delivered late, in bulk, once the tab was focused
again. CDP eval keeps answering throughout because the debugger drives the renderer outside
the ordinary task queues.

Two things work correctly and are worth stating: **closing** the holder promotes a successor
as designed (`init` 333 ms via cache hit, then local answers), and focusing the holder again
repairs the other tab instantly, without a reload. Only backgrounding is affected, and only
when two or more tabs are open — a single tab is its own holder and is focused whenever it is
used.

Reproduced twice, on two different tab pairs (A→B and B→C).

Not checked: desktop Chrome against this branch; whether the service worker (absent from the
probe stand) changes anything; power-saving modes; and the case where the relay, not the
holder, is backgrounded.

resources/public/js/sqlite3-dictionary.js
```diff
@@ -0,0 +1,266 @@
...
+// precisely so that the tab holding it is one the system will not freeze.
+//
+// `visibilitychange` fires the moment a tab goes away and the freeze lands
+// 50-60 s later, so the handover has three orders of magnitude more time than
+// it needs. If a tab is killed outright instead, the browser releases its lock
+// and the next tab takes over — which is the same path as closing one.
+//
+// Known limitation, accepted: two tabs visible at once — split screen, or two
+// windows side by side — still get one dictionary between them. The one that
+// took the lock answers; the other waits, exactly as if it were hidden.
```
u473t8: Разве даже в этом случае не активна всего лишь одна вкладка, где пользователь печатает?

resources/public/js/sqlite3-dictionary.js
```diff
@@ -0,0 +1,266 @@
...
+
+const SAH_POOL_LOCK = "sprecha-sqlite-opfs-sahpool";
```
u473t8: Никаких упоминаний "sprecha" в коде

resources/public/js/sqlite3-dictionary.js
```diff
@@ -0,0 +1,266 @@
...
+
+// Where this worker is, as the message its page is sent. `loading` covers both
+// "starting up" and "waiting for another tab to finish", because from the
+// page's side they are the same thing: no dictionary yet, ask anyway.
+let lifecycle = { type: "loading" };
+
+// Requests with nowhere to go yet. Never dropped: before #351 a query asked
+// early answered with an empty list, which looks exactly like a word the
+// dictionary does not have.
+let waiting = [];
```
u473t8: Я не понял, откуда у нас очередь реквестов, если вкладка, владеющая базой, общается с ней напрямую?

adr/0012-dictionary-holder-elected-by-web-lock.md
u473t8: That ADR is gone from the branch — the holder scheme it described was replaced, and the file is now `adr/0012-dictionary-belongs-to-the-visible-tab.md`.

In the current design there is no holder to go away. A tab owns the database only while it is visible and gives it back on `visibilitychange`, before the platform can freeze it. Queries asked while a tab has no turn are held in `waiting` and answered when its turn comes — nothing is dropped, and nothing crosses a tab boundary. If a tab is killed without running its handler, the browser releases the lock and the next tab takes over; that is the same path as closing one, and it is covered by a test.

openspec/changes/archive/2026-08-16-one-dictionary-for-every-tab/design.md
u473t8: Yes, and that is what the change became — ownership follows visibility, so the tab you are looking at holds the database. This file was the earlier holder-and-relay design and no longer exists in the branch.

openspec/changes/archive/2026-08-16-one-dictionary-for-every-tab/proposal.md
u473t8: `ifAvailable: true` asks the lock manager to fail immediately rather than queue. The second tab therefore got `null`, posted `another-tab-open` and never tried again, so its suggestion list stayed empty for as long as the tab was open — with nothing in the UI to say why.

It is gone now. The request queues, so a tab that cannot have the database yet simply waits its turn instead of being turned away permanently.

resources/public/js/sqlite3-dictionary.js
u473t8: Верно, печатают всегда в одной. Но критерий сейчас — `document.hidden`, а он у обеих видимых вкладок `false`, поэтому обе считают себя вправе, и базу получает та, что первой взяла лок. Если печатают в другой, она ждёт.

Различить можно по фокусу: `document.hasFocus()` плюс `focus`/`blur` на window — тогда база уходит именно туда, куда печатают, и ограничение снимается почти полностью. Цена — более частые передачи по ~14 мс: фокус теряется не только при переходе в другую вкладку, но и при клике в адресную строку, в devtools или в другое приложение, а видимость в этих случаях не меняется. Могу сделать, если считаешь нужным.

resources/public/js/sqlite3-dictionary.js
u473t8: Это единственное место, где бренд попал в идентификатор. В остальном `Sprecha` встречается только как имя продукта — `manifest.json`, логотип в `application.cljs`, заголовок в `install_guide/view.cljs`.

Префикс тут ничего не даёт: Web Lock и так изолирован origin'ом, никакого глобального пространства имён у него нет. Можно сократить до `sqlite-opfs-sahpool`.

Единственный риск при переименовании — если старая и новая версии работают одновременно, они возьмут разные локи и обе полезут в один файл OPFS. У нас этого не будет: service worker без `skipWaiting`, новая версия активируется только когда закрыты все вкладки старой. Переименовать?

resources/public/js/sqlite3-dictionary.js
u473t8: Очередь не межтабовая — она внутри одной вкладки, между её страницей и её же воркером.

Вкладка может получить запрос раньше, чем у неё появится база: пока ждёт своей очереди на лок, пока грузится wasm при первом ходе, пока открывается файл. Такой запрос кладётся в `waiting` и отвечается, когда ход наступит. До #351 он получил бы `[]`, что для пользователя неотличимо от «такого слова нет» — это и была исходная жалоба.

Когда база у вкладки есть, `request()` отвечает сразу, и очередь не участвует вовсе.

resources/public/js/sqlite3-dictionary.js
```diff
@@ -0,0 +1,315 @@
...
+  try {
+    await measure("pause", () => {
+      closing.close();
+      pool.pauseVfs();
+    });
+  } catch (ignored) { /* the turn ends either way */ }
+}
+
+
+function answer(msg) {
```
u473t8: This function should take db as an argument. Can we stick to data driven design here?

resources/public/js/sqlite3-dictionary.js
```diff
@@ -0,0 +1,315 @@
...
+
+
+// The response to one request: given straight back if this tab has a turn,
+// otherwise awaited until it gets one and asked again. Never dropped —
+// answering early with an empty list looks exactly like a word the dictionary
+// does not have.
+async function request(msg) {
+  const now = respond(msg);
+  if (now) return now;
+  await whenSettled();
```
u473t8: I still wonder what terrible from the UX perspective would happen if we do not wait for the settlement and just return empty response.

src/client/instrumentation.cljs
```diff
@@ -114,20 +114,34 @@
...
+   `:phases` carries what each later turn cost."
   []
-  (let [^js measure (try
-                      (.measure js/performance "dictionary-ready" "dictionary-start")
-                      (catch :default _ nil))]
-    (when measure
-      (swap! metrics assoc-in [:dictionary :ready-ms] (.-duration measure)))))
+  (when-not (get-in @metrics [:dictionary :ready-ms])
+    (let [^js measure (try
+                        (.measure js/performance "dictionary-ready" "dictionary-start")
```
u473t8, line 129: Как работают аргументы в этой функции?

resources/public/js/sqlite3-dictionary.js
u473t8: Done in 0231219.

`answer(db, msg)` and `respond(db, lifecycle, msg)` now take the state as arguments instead of reaching into module scope. `request` is the only function left that reads the module state, and it reads it once and passes it down, so everything below it is a decision over its arguments.

resources/public/js/sqlite3-dictionary.js
u473t8: Nothing terrible on a warm start, and the queue is gone as of 0231219 — the empty answer is what a tab without a turn gives now.

Counted against the numbers already in the tree. The waiting windows are ~14 ms to take the lock back from the tab in front and ~94 ms for a tab's first turn. In front of the query sits a 100 ms trailing debounce, and typing runs 120-180 ms per key (#195). So on a warm start the turn has begun before the first query is even sent: the user types **zero** characters into that window. On a page load it is not close — the worker is constructed in `init!`, so the 94 ms run against app boot and first render, long before anyone reaches the field.

The one window that is real is a cold start, where the dictionary is still downloading. That is seconds, and at 120-180 ms per key the user types the whole word and the next one into it. There an empty list does read as "the dictionary does not have this word" — but holding queries buys that single case and costs a mechanism everywhere, so it is accepted for now and revisited as its own task if it turns out to grate.

Two things kept so the case stays addressable:

- an open failure is still returned as an error, never flattened into an empty list, so a broken dictionary stays distinguishable from a missing word;
- `:dictionary/ready?` stays on the port as a reading (not a gate), which is what lets a caller tell "no completions for this prefix" from "no dictionary in this tab" — #312 needs exactly that.

Not measured: cold start on the shipped dictionary and behaviour on a phone. The suite runs the ~28 KB fixture on localhost, so the user-facing cost of this choice is reasoned, not observed.

src/client/instrumentation.cljs
u473t8: `performance.measure(name, startMark, endMark)`. Здесь третий аргумент опущен, а при опущенном конце спецификация берёт момент вызова — то есть меряется отрезок от метки до «сейчас». Возвращается объект `PerformanceMeasure`, отсюда `.-duration`.

Метка ровно одна: `dictionary-start`, её ставит `dictionary-start!` из `init!` в `db/sqlite.cljs`, перед конструированием воркера. Второй метки нет вообще — концом служит момент, когда со стороны воркера приходит `ready`. Метка живёт на таймлайне документа намеренно: у воркера свой time origin, и его `performance.now` пришлось бы приводить арифметикой, чтобы сравнить с этим.

`try/catch` нужен ради одного случая: если метки `dictionary-start` в буфере нет, `measure` бросает `SyntaxError`, а не возвращает nil.

`dictionary-phase!` к этому API отношения не имеет — фазы воркер меряет сам своим `performance.now()`, сюда приходят готовые `durationMs`.

Отдельно стоит держать в голове, что все три вызова стоят под `goog/DEBUG`: в релизной сборке ни меток, ни измерений нет — это #298 и #311.

src/client/db/sqlite.cljs
```diff
@@ -32,46 +39,99 @@
...
+  "Where the worker is: it takes the database on every turn and gives it back
+   when the tab leaves the foreground, and says so each time.
+
+   Queries are sent whatever it says — nothing here gates them, because a
+   worker without the database answers empty by itself. What is kept is the
+   reading `ready?` returns, for a caller that has to describe the difference
+   (#312); the rest feeds the measurement and the log."
+  [holding? ^js e]
+  (case (.. e -data -type)
+    "loading" (reset! holding? false)
```
u473t8: It is a bit stupid to recieve atom as an argument, and then modify it.

src/client/db/sqlite.cljs
```diff
@@ -32,46 +39,99 @@
...
+
+(defn attach
+  "Builds the `:db/sqlite` component around a worker. Split out of `init!` so
+   the protocol can be driven from a test, in a runtime with no workers to
+   spawn."
+  [worker]
+  ;; Two atoms because they answer to two different things: `requests` is the
+  ;; id counter the proxy needs and nobody outside it reads, `holding?` is the
+  ;; reading `ready?` hands out.
+  (let [requests (atom {:next-id 0})
```
u473t8: WHy do we need track requests?

src/client/db/sqlite.cljs
u473t8: Done in 5d8e5a1.

Split the decision from the effect. `holding-state` is now a pure function of one message — `true` when the tab took its turn, `false` when it gave it back or could not open the database, `nil` when the message is not about ownership — and it touches nothing. `attach` applies the answer to the atom it owns.

The log and `instrumentation/*` moved out with it, into `observe-worker-message!`, so parsing is no longer interleaved with reporting.

That makes the parse testable without any mock: `holding-state-reads-one-message` states the protocol table directly against `#js` literals, with no stub worker involved. It is public for exactly that reason. The wiring stays covered separately, through `attach` and `ready?`.

One detail worth naming: `attach` uses `when-some`, not `when-let` — `false` is a real answer here, and only `nil` means "not about ownership".

src/client/db/sqlite.cljs
u473t8: Good question, and the honest answer turned out to be: we do not. Removed in 5d8e5a1.

What it was for: `postMessage` gives no reply channel, so a response had to be paired with its request some other way. Several queries can be in flight — the suggest effect debounces at 100 ms but does not cancel — and the worker's lifecycle broadcasts (`loading`, `ready`, `error`, `phase`) arrived on the same listener, so each request also had to learn to skip those. The id was how both problems were solved.

Your `MessageChannel` suggestion removes the need for either. A query now carries a fresh channel, the worker answers on `e.ports[0]`, and the reply reaches exactly the promise that asked. Gone with the counter: the id field on both sides, the per-request `addEventListener`/`removeEventListener` on the shared port, and the `id`-filter in the response handler. `answer` and `respond` in the worker dropped `{id: msg.id, ...}` too.

It also separates the two conversations that were sharing one pipe: the worker's own port now carries only broadcasts — news for the tab — and each query is answered on its own channel. A reply can no longer be confused with a broadcast, which is a whole class of bug that had a test to itself.

On the cost, since a channel per keystroke sounds expensive — measured rather than assumed, 20k round trips per mode against an echo worker in Chrome 144:

| transport | per round trip |
|---|---|
| id correlation | 0.103–0.119 ms |
| `MessageChannel` | 0.137–0.153 ms |

So ~34 µs dearer per query. That sits behind a 100 ms trailing debounce and in front of a query that costs 20–27 ms (#179) — about 0.15% of one lookup, at a rate the debounce caps near 10/s. Not a price worth keeping a counter to avoid.

Node has `MessageChannel` even though it has no `Worker`, so the unit tests drive real ports rather than a fake: `two-queries-in-flight-get-their-own-answers` answers two queries out of order and pins that nothing about ordering is what pairs a reply with its query.

Both suites green after the change: node 192/457, browser 26/26.

u473t8 (commented):
LGTM

resources/public/js/sqlite3-worker.js
```diff
@@ -1,95 +1,27 @@
...
-  try {
-    self.postMessage({ id: msg.id, result: db.exec(msg) });
-  } catch (err) {
-    self.postMessage({ id: msg.id, error: String(err) });
+// The page's end of the dictionary. Every tab has one of these and they never
+// speak to each other; which of them has the database open is decided by the
+// browser's lock manager, and one without the lock answers with no
+// completions rather than queueing the query.
+//
+// Two channels, on purpose. Where this worker is — loading, ready, error,
```
u473t8: What two channels?

u473t8: Fair — the wording promised a matched pair, and there is no pair. Rewritten in 9ffff08.

There are two kinds of traffic and they are not alike:

- **standing and shared** — the worker's own port, open for the tab's whole life, carrying only where the worker is: `loading`, `ready`, `error`, `phase`. News for the tab, not a reply to anyone.
- **per query and disposable** — the page makes a fresh `MessageChannel` for each query, keeps `port1`, sends `port2` with the message, and the answer comes back down that one before it is closed. It lives exactly as long as the one answer it carries.

The reason for keeping them apart is unchanged: a reply cannot be mistaken for a broadcast, and two queries in flight cannot be mistaken for each other, because the pairing *is* the channel — no id to match, no counter to keep.

While checking whether the same imprecision was anywhere else, I found a worse one: `openspec/specs/sqlite-dictionary-worker` still described the transport as "Promises resolved by matching request id", which 5d8e5a1 removed. That is now restated to match the code, in the live spec and in the archived change's delta together. Nothing else claimed "two channels".

Comments and spec prose only — `git diff` has zero non-comment lines in the JS.

On testing this: the node suite never loads `sqlite3-worker.js`, so it proves nothing here. The browser suite is the only run with any power, since it is what actually executes the file — worth it only to catch a malformed comment block, which `node --check` also settles. Ran all three anyway: `node --check` clean on both worker scripts, node 192/457, browser 26/26.

Issue #365, opened 2026-08-17
Keep a translation exactly as it was typed


u473t8:
A translation is entered as one line and split on `[,;.]` (`domain/vocabulary.cljs`,
`parse-translations`). The separator lives inside the data: `без того, чтобы` becomes
two translations, and so does any aspect pair written with a dot.

## The split buys nothing

- Grading never looks at the elements. `vocab->trial` (`domain/lesson.cljs`) puts the
  translations into `:prompt` and the German `(:value vocab)` into `:answer`; the typed
  answer is compared against `:answer` only.
- Display glues the vector back together with `str/join ", "` — the same separator it
  was cut on.
- The elements are used for one thing: deduplication in `merge-translations` when the
  same word is added again.

The string is cut on a comma so it can be joined by a comma again, and the only
observable effect is that translations containing a comma are destroyed.

## Proposal

- `parse-translations` keeps the entered text as a single translation.
- The field accepts several lines, for readability.
- `Enter` inserts a newline instead of submitting; `Ctrl`/`Cmd`+`Enter` submits.
- The submit button stays and is the only path on a phone, where neither modifier can
  be typed.
- Both `str/join ", "` sites in `pages/home/actions.cljs` (suggestion prefill and picked
  suggestion) hand over the text as stored.

The document schema does not change: `:translation` stays a vector, normally holding one
element, and existing documents are read as they are. `update-word` shares
`parse-translations`, so editing follows the same rule.

## Already in place

The field is a `textarea`. `submit-if-enter` already ignores `Enter` while `Shift` is
held, so a newline can be typed today — it is only torn apart on save. On a phone `Enter`
inserts a newline by default and submission goes through the button.

`Enter` in the German word field has a different meaning — pick a suggestion, or move to
the translation field when there are none (`handler-word-keydown`). Untouched.

Examples already store their translation as one string (`:prompt (:translation example)`
in `example->trial`), so the unsplit shape is present in the code today.

## Given up knowingly

Deduplication on re-adding a word gets coarser: whole strings are compared instead of
variants inside them. If a drill that accepts any one of the translations appears, or a
display that shows them one at a time, the elements will have to be recovered. Neither
exists today.

## Acceptance

- A translation containing a comma, a dot or a slash is stored exactly as entered.
- Several lines are stored and shown as entered.
- Submitting works on a phone without a hardware keyboard.

The other half of this is #362: the dictionary transport still splits on a comma.

PR #368, opened 2026-08-17
GH-362 carry completions' translations as a JSON array

### LINKED ISSUE #362: Translations containing a comma are split into pieces
`completions-sql` aggregates translations with
`GROUP_CONCAT(DISTINCT t.value ORDER BY t.rank ASC)` and the client splits the result
on `,` (`adapters/dictionary.cljs`). The separator is not distinguishable from the
content: **1100 of 301 683 translations contain a comma themselves**, so they come
back split in the wrong places.

Measured against the shipped dictionary:

| lemma | stored translation | what the client shows |
|---|---|---|
| `ohne` | `без того, чтобы` | `без того` + ` чтобы` |
| `eigentlich` | `действительный, настоящий` | two entries |
| `unten` | `внизу, снизу` | two entries |
| `zwar` | `хотя..,но..` | two entries |

These are frequent words, so the damage is not in the tail. A wrong split is then
prefilled into the translation field and saved, so it reaches the user's vocabulary
and their reviews.

## Scope

Make the transport unambiguous rather than picking a rarer separator: `json_group_array`
returns a real array and the client parses it, which cannot collide with content. Any
other fix has to answer what happens the day a translation contains the new separator.

## Acceptance

- A translation containing a comma arrives whole.
- Multiple translations still arrive as separate entries, ordered by rank.
- No change to the query's cost on short prefixes (#179 measured 20-27 ms for `f`).

u473t8:
Closes #362.

## The defect

`completions-sql` joined a lemma's translations with `GROUP_CONCAT` and the client split the result on `,`. The separator is also content — 1100 of 301 683 translations contain a comma — so every one of them was torn into pieces, and a lemma with no translations got one blank.

Diffing the old query decoded the old way against the new one, over whole result sets from the shipped dictionary:

| lemma | before | after |
|---|---|---|
| `unten` | `внизу` / ` снизу` / `ниже` | `внизу, снизу` / `ниже` |
| `eigentlich` | `действительный` / ` настоящий` / `собственный` | `действительный, настоящий` / `собственный` |
| `die Hose` | `брюки` / `штаны` / `смерч ` / ` (небольшое) торнадо` | `брюки` / `штаны` / `смерч , (небольшое) торнадо` |
| `unterhalten` | one blank translation | none |

23 rows compared: 19 byte-identical, 0 reordered, 4 repaired.

## Why `json_group_array`

`GROUP_CONCAT(DISTINCT value, char(31))` does not compile — *"DISTINCT aggregates must have exactly one argument"*. That leaves a separator plus a dedupe subquery, or a JSON array.

The separator variant needs a character reserved forever. `char(31)` is free across all 301 683 translations today, but nothing in the build pipeline enforces that, so the claim has to be re-verified on every rebuild by somebody who remembers to. A JSON array has no separator, so there is no invariant to keep. It also measured cheaper than the subquery.

## Cost

#179 fixed this query's shape so only ten surviving lemmas reach the translation lookup. That is exactly the part this change touches, so it was re-measured: shipped 37 MB dictionary, the vendored SQLite 3.53.1 WASM build, its own `opfs-sahpool` VFS in a real browser, variants interleaved inside each repetition, medians of 61.

| prefix | `GROUP_CONCAT` (before) | `char(31)` + subquery | `json_group_array` (after) |
|--------|------------------------:|----------------------:|---------------------------:|
| `f`    | 47.5 ms | 48.0 ms | 45.4 ms |
| `fe`   | 7.5 ms  | 7.7 ms  | 7.6 ms  |
| `fen`  | 1.5 ms  | 1.5 ms  | 1.6 ms  |
| `sch`  | 31.0 ms | 31.2 ms | 30.7 ms |
| `hau`  | 4.0 ms  | 4.6 ms  | 4.4 ms  |
| `a`    | 83.2 ms | 78.9 ms | 82.8 ms |

End to end the three are indistinguishable — every difference is smaller than the run-to-run spread and its sign is not stable between runs, because the prefix range scan dominates and is identical in all three.

Isolating the aggregate for a fixed set of ten lemma ids (batches of 200 execs, median of 25 batches) separates it from that noise:

| variant | per call | vs before |
|---------|---------:|----------:|
| `GROUP_CONCAT` | 0.329 ms | — |
| `char(31)` + subquery | 0.429 ms | +30 % |
| `json_group_array` | 0.354 ms | +8 % |

+0.025 ms per completions call, 0.03 % of the one-letter prefix. The rejected alternative is the expensive one.

The absolute figures are higher than #179's 20-27 ms for `f` — different hardware, and a live app tab in the same browser. Only the before/after within one run is comparable.

## Not in scope

The input half. The home form still prefills with `(str/join ", " translations)` and `parse-translations` splits that again on `[,;.]`; that round trip is #365's subject. This change makes the dictionary side honest so #365 has something intact to store.

## Verification

- `npx shadow-cljs compile node-test` — 193 tests, 448 assertions, 0 failures. New `test/client/adapters/dictionary_test.cljs` covers the comma case, order, the empty lemma, a null cell, prefix normalization, and the two paths that never reach the worker. Confirmed the suite fails when an expectation is flipped, so the async assertions really run.
- The shipped SQL string, extracted from the source, run against the shipped dictionary through the real WASM build and OPFS pool — the table above.

Not verified: no browser click-through of the add-word form, and no measurement on the owner's phone.

### DIFF
--- a/src/client/adapters/dictionary.cljs
+++ b/src/client/adapters/dictionary.cljs
@@ -1,6 +1,5 @@
 (ns adapters.dictionary
   (:require
-   [clojure.string :as str]
    [db.sqlite :as sqlite]
    [utils :as utils]))
 
@@ -17,7 +16,16 @@
    for the surviving ten alone. The previous shape joined and GROUP_CONCATed
    every in-range lemma — thousands on a one-letter prefix — and discarded all
    but ten after sorting, which is why short prefixes cost ~100x more.
-   Measured in #179: prefix f 95-119 ms -> 20-27 ms."
+   Measured in #179: prefix f 95-119 ms -> 20-27 ms.
+
+   Translations travel as a JSON array, not a joined string. GROUP_CONCAT's
+   separator is also content: 1100 of the dictionary's translations contain a
+   comma, so splitting the join tore `тем, что` into `тем` and ` что` and gave
+   `indem` five translations where it has three. A reserved separator would work
+   only as long as nobody's rebuild introduces the character, and nothing
+   enforces that; an array has no separator to reserve. Measured against the
+   shipped dictionary in #362: the aggregate costs +0.025 ms per call, which is
+   0.03% of the one-letter prefix, and the #179 shape above is untouched."
   "WITH top AS (
      SELECT l.id, l.value, l.pos, l.rank
      FROM (SELECT DISTINCT lemma_id
@@ -35,7 +43,7 @@
      EXISTS (SELECT 1
              FROM surface_forms sf
              WHERE sf.normalized_form = ? AND sf.lemma_id = top.id) AS has_exact,
-     (SELECT GROUP_CONCAT(DISTINCT t.value ORDER BY t.rank ASC)
+     (SELECT json_group_array(DISTINCT t.value ORDER BY t.rank ASC)
       FROM translations t
       WHERE t.lemma_id = top.id) AS translations
    FROM top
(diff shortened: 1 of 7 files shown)


u473t8 (commented):
LGTM

src/client/adapters/dictionary.cljs
```diff
@@ -64,9 +84,5 @@
                                              :rowMode     "object"}))
                           :keywordize-keys
                           true)]
-          (for [{:keys [lemma translations has_exact pos]} rows]
-            {:lemma        lemma
-             :pos          pos
-             :translations (str/split (or translations "") #",")
-             :exact?       (pos? has_exact)}))))
+          (map completion rows))))
```
u473t8: No point of extracting completion.

PR #369, opened 2026-08-17
GH-365 keep a translation exactly as it was typed

### LINKED ISSUE #365: Keep a translation exactly as it was typed

u473t8:
Closes #365.

`parse-translations` split the entered text on `[,;.]`, so `без того, чтобы` became two
translations. Nothing read the pieces — grading compares the typed answer against the German
`:answer`, display glued them back with the same separator — so the split's only observable
effect was the damage. Phrases already stored their translation whole; words now do too.

## What changed

- `parse-translations` keeps the entered text as one entry; blank still yields none, so the
  `:empty-translations` guard holds. `update-word` shares it, so editing follows the same rule.
- The translation field takes several lines. `Enter` inserts a newline; `Ctrl`/`Cmd`+`Enter`
  submits, through the shared `:action/submit-if-ctrl-enter` (now `Cmd`-aware and
  default-suppressing, so no browser can submit twice). The button is untouched and stays the
  only path on a phone. `Enter` on the German input keeps its own meaning.
- The dictionary prefill trims the pieces the transport's split left before rejoining them:
  a stored `без того, чтобы` no longer gains a second space, and `пёс, собака` still reads as
  `пёс, собака`. #362 removes the split at the source.
- The word edit dialog's translation field becomes a `textarea` like the phrase one — an
  `input` silently swallows the line breaks it is handed, which would have thrown away a
  multi-line translation on the next save. Its autogrow rules moved from the `--phrase`
  modifier to a `--multiline` one that both kinds carry.
- `white-space: pre-line` on the word-list translation and the lesson prompt, so a stored line
  break is shown as one.

The document schema does not change: `:translation` stays a vector, normally holding one
element, and documents written before this are read as they are.

## Given up knowingly

Deduplication on re-adding a word compares whole strings instead of variants inside them. If a
drill that accepts any one of the translations appears, or a display that shows them one at a
time, the elements have to be recovered. Neither exists today.

## Verified

Node tests green (`npx shadow-cljs compile node-test`). New cases: a comma, `;`, `.` and `/`
survive; several lines survive; blank yields no entries; the edit path inherits the rule; a
document with several stored entries is left alone; the prefill reconstructs
`без того, чтобы`.

In a real browser, against a build of this branch served on a scratch backend, driven through
CDP with real key events:

| Checked | Result |
|---|---|
| `Enter` in the translation field | inserts a newline, form not submitted (submit count 0) |
| `Ctrl`+`Enter` | submits exactly once |
| `Cmd`+`Enter` | submits exactly once |
| `ДОБАВИТЬ` button | submits |
| `Enter` on the German input | still moves focus to the translation field |
| Stored document | one entry, `без того, чтобы\nвторой вариант` — comma and newline intact |
| `первый; второй. третий/четвёртый` | stored as one entry |
| Edit dialog | opens as a `textarea` holding both lines; `Ctrl`+`Enter` saves; the saved entry keeps its comma |
| Phrase edit dialog | unchanged geometry after the CSS split (column, full width, `field-sizing: content`, 48–144 px) |

add form
word list
edit dialog

## Not proven

- A real phone. The mobile path — `Enter` inserting a newline on a soft keyboard, the button
  being the only way to submit — was checked only in a desktop browser at a 412×800 mobile
  viewport, not on hardware.
- Safari. `field-sizing: content` and the autogrow effect were exercised in Chrome only.

### DIFF
--- a/openspec/changes/archive/2026-08-17-keep-typed-translation-whole/.openspec.yaml
+++ b/openspec/changes/archive/2026-08-17-keep-typed-translation-whole/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven-with-adr
+created: 2026-08-17
--- a/openspec/changes/archive/2026-08-17-keep-typed-translation-whole/proposal.md
+++ b/openspec/changes/archive/2026-08-17-keep-typed-translation-whole/proposal.md
@@ -1,0 +1,66 @@
+## Why
+
+A translation entered on the home add-word form is split on `[,;.]` before it is stored
+(`domain/vocabulary.cljs`, `parse-translations`), so the separator lives inside the data:
+`без того, чтобы` becomes two translations, and so does any aspect pair written with a dot.
+
+The split buys nothing. Grading compares the typed answer against the German `:answer`
+(`domain/lesson.cljs`, `vocab->trial`), never against the elements; display glues the vector
+back together with the same separator it was cut on. The elements are read for one thing —
+deduplication in `merge-translations` when the same word is added again. The string is cut on
+a comma so it can be joined by a comma again, and the only observable effect is that
+translations containing a comma are destroyed. Phrases already store their translation whole
+(`domain/phrase.cljs`), so words are the outlier.
+
+## What Changes
+
+- `parse-translations` keeps the entered text as one translation entry. A blank entry still
+  yields no translations, so the `:empty-translations` guard in `use-cases/vocabulary` holds.
+- The translation field accepts several lines. `Ctrl`/`Cmd`+`Enter` submits; bare `Enter`
+  inserts a newline. The submit button stays and remains the only path on a phone, where
+  neither modifier can be typed.
+- The dictionary prefill hands over the stored text rather than re-spacing it: both
+  `str/join ", "` sites in `pages/home/actions.cljs` trim the pieces the transport's split
+  left, so a stored `без того, чтобы` reaches the field as it is stored instead of gaining a
+  second space, while several translations still read as `пёс, собака`.
+- The word edit dialog's translation field becomes a `textarea` like the phrase one: an
+  `input` silently swallows the line breaks a multi-line translation carries.
+- A stored line break is shown as one: `white-space: pre-line` on the word list's translation
+  and on the lesson prompt.
+- `Enter` on the German word input is untouched — it picks the highlighted suggestion or moves
+  to the translation field (`handler-word-keydown`).
+- The document schema does not change: `:translation` stays a vector, normally holding one
+  element, and existing documents are read as they are. `update-word` shares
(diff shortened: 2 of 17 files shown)


u473t8 (commented):
Из скриншотов не видно, форматируется ли перевод как список, когда мы вводим его через enter. Опять же, что будет с отображ нием слов в списке слов. Нужно походу пересмотреть и не показывать перевод в эоементе списка, показывать по клику в модалке. То есть тап по элемент открывает модалку с редаткируемым полем перевода.

u473t8 (commented):
LGTM

PR #370, opened 2026-08-17
GH-358 scope the picked mode to the picked lemma

### LINKED ISSUE #358: Editing a picked suggestion does not re-evaluate the phrase heuristic
Pick a word from the suggestions, then keep typing to turn it into a phrase — the
form stays in word mode. The phrase heuristic does not run again on the longer
value, so the entry is saved as a word.

Steps: type `das`, pick `das Haus` from the list, append ` ist gross`. Expected the
form to switch to phrase mode; it stays a word.

## Where to look

`domain/phrase.cljs`: `add-mode` is `(or override (if (phrase-value? …) :phrase :word))`,
so an explicit override always wins. Picking a suggestion is what sets that override
(`pages/home/actions.cljs`, `select-item`), and nothing clears it when the value is
edited afterwards. Worth deciding whether the override should survive any edit at all,
or only until the value stops matching what was picked.

## Acceptance

- Editing the value after picking a suggestion re-evaluates the mode.
- Picking a suggestion and submitting unchanged still respects the picked lemma.

u473t8:
Closes #358.

Picking a suggestion set `:home/mode-override` and nothing ever cleared it, so
typing on from a picked word kept the form in word mode and the entry went to
the database as a word.

The override could not simply be dropped: it is what makes `das heißt` a
phrase, which the article exception reads as a word, and the suggestion list it
came from is gone by then. So it is scoped instead — the pick decides for the
lemma it was made for and stops deciding once the value is something else.
Sameness is the vocabulary identity already in use (`normalize-value`), so
case, punctuation and a trailing space do not revoke a pick; they do not make a
different document either.

A pick is still the only source of an override — the form has no manual mode
control.

## Verification

- `npx shadow-cljs compile node-test` — 188 tests, 445 assertions, 0 failures.
  New: `add-mode` unit tests (pick held on its own lemma, dropped after an
  edit, `das heißt` unchanged) and two home-page tests, one on the legend the
  presenter produces, one on which use case the submit reaches.
- `npx playwright test --project=desktop` — 18 passed, including the new
  `test/browser/picked-suggestion-mode.spec.js`: type `Haus`, pick `das Haus`,
  append ` ist gross`, panel title turns into "Добавить фразу".
- Mutation-checked both ways: with the old rule restored, the three new unit
  assertions and the browser spec fail; with the fix, all green.

Before the append (picked lemma, word mode) and after:

picked word
typed on into a phrase

Not covered: the mobile Playwright project was not run (this change touches no
layout), and nothing here was tried on a phone.

### DIFF
--- a/openspec/changes/archive/2026-08-17-rescope-pick-mode-override/.openspec.yaml
+++ b/openspec/changes/archive/2026-08-17-rescope-pick-mode-override/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven-with-adr
+created: 2026-08-17
--- a/openspec/changes/archive/2026-08-17-rescope-pick-mode-override/proposal.md
+++ b/openspec/changes/archive/2026-08-17-rescope-pick-mode-override/proposal.md
@@ -1,0 +1,41 @@
+## Why
+
+Picking a suggestion sets a mode override that nothing ever clears, so typing on
+from the picked lemma keeps the form in the mode the pick chose: `das`, pick
+`das Haus`, append ` ist gross`, and a phrase is saved as a word (GH-358).
+
+The override cannot simply be dropped — it is what makes `das heißt` a phrase,
+which the space heuristic reads as an article pair — so it has to be scoped
+instead of removed.
+
+## What Changes
+
+- A pick's mode applies only while the entered value still is the lemma that was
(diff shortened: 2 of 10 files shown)


u473t8 (commented):
LGTM

Issue #349, opened 2026-08-15
Phone suggestion list covers the translation field


u473t8:
On a phone the suggestion list opens as an overlay above the rest of the add
form (#289, PR #291). That removed the layout shift it used to cause — the list
no longer pushes the value field out from under the finger — but it traded the
shift for occlusion: while suggestions are visible they cover the translation
field, and the top edge of `ДОБАВИТЬ`.

Why it matters: the two fields are filled in sequence, word then translation.
The list stays open until the value field is committed or blurred
(`:action/dismiss-suggestions` on blur, `src/client/pages/home/view.cljs`), so
during the whole of the first step the target of the second step is hidden. The
user cannot see what they are about to type into, and a tap aimed at the
translation field lands on a suggestion instead.

This is the standard autocomplete trade and it is what the desktop layout
already does; at 1280x720 there is room below the panel, so nothing is
occluded in practice. At 390x844 there is not.

## Scope

Decide and implement how the phone shows suggestions without either shifting
the form or hiding the next field. Options worth measuring against each other:

- Shorten the list on phones so it stops above the translation field.
- Anchor it below the whole panel rather than directly under the value field.
- Show it only while the value field is focused and empty of a chosen match,
  dismissing on the first suggestion tap.
- Give it its own sheet, the way a native picker does.

## Acceptance

- With the suggestion list open at 390x844, the translation field and the
  submit button are fully visible.
- The value field, the panel title and the submit button still do not move
  while a word is typed — the assertions in
  `test/browser/add-form-stability.mobile.spec.js` continue to hold.

Related: #289 introduced the overlay, #292 is the keyboard bounce on submit,
#340 is the wider mobile layout stabilisation, #321 is the same list's screen
reader support.

Issue #289, opened 2026-08-12
Assert layout stability of the add form on a mobile Playwright project


u473t8:
Phase 2 of #177 never landed. `src/client/instrumentation.cljs` exposes `window.__metrics()` under `goog.DEBUG` (renders, dispatches, frames, layout shifts, long tasks), the Playwright suite runs in CI, and nothing connects them: `test/` holds no reference to `__metrics`.

Two gaps make a naive assertion useless:

- The layout-shift counter filters entries with `hadRecentInput`, the CLS rule. Typing is recent input, so shifts that happen while the user types — the ones that prompted the GH-285 fixes — never reach the counter.
- `playwright.config.js` sets no viewport, so the phone media queries (`hover: none`, `pointer: coarse`, `max-width: 600px`) never apply. The in-flow suggestion list and everything it pushes are untested at any width.

Scope:

- Add an unfiltered shift counter beside the CLS-filtered one, leaving the existing one as is.
- Add a `mobile` project (390x844, `hasTouch`) with a seeded dictionary, since suggestions need one to appear.
- Assert on typing a word into the add form: unfiltered shift score below 0.05, `y` of the value field and of the panel title unchanged, at most two renders per keystroke, no long tasks.

Measured for reference on the desktop stand after the GH-285 fixes: the value field and the panel title hold still, while the submit button moves ~184px when the suggestion list opens in flow. Decide as part of this work whether that push is acceptable or whether the list should overlay on phones too.

`test/browser/add-form-stability.spec.js` covers the desktop half already; this issue is the phone half.

Needs: #261

PR #374, opened 2026-08-18
GH-373 phone suggestion list back in flow, height by content

### LINKED ISSUE #373: Suggestion list does not shrink to one or two rows on a phone
Reported from the installed app: when the suggestion list is down to one or two rows,
the box does not shrink to fit them.

## Not reproduced on desktop — measured, both statically and over time

Chrome 151 against production, real dictionary. Static, by number of rows:

| rows | box height |
|---:|---:|
| 1 | 49 px |
| 4 | 169 px |
| 5 | 209 px |
| 10 | 240 px (the cap) |

Dynamically, going from a one-letter prefix to a long one:

```
settled on "f"        10 rows   240 px
+0 ms                 10 rows   240 px
+50 ms                10 rows   240 px
+100 ms                1 row     49 px
+1200 ms               1 row     49 px
```

The 100 ms step is the suggest debounce, and the list keeps its previous answer until
the next arrives on purpose (#178). Past that it shrinks immediately. `max-height` is
the only height rule in `resources/public/css/components/autocomplete.css`; nothing sets
a fixed height, and no other stylesheet mentions the list.

So the desktop path is clean and the defect lives somewhere the desktop measurement does
not reach.

## Where to look

- The phone branch caps at `min(240px, 34svh)`. `svh` is the *small* viewport height and
  knows nothing about the software keyboard, so with the keyboard up the list can occupy
  more of the visible area than intended — which looks like "not shrinking" even when the
  box does fit its rows.
- With ten rows of 40 px the content is 400 px against a 240 px cap, so the list scrolls.
  A learner who has typed most of a word may still be getting ten lemmas sharing that
  prefix (`Fenster`, `Fensterbank`, `Fensterbrett`…) rather than the one they expect, and
  reads a full-height scrolling box as a box that refuses to shrink.
- #349 covers the list covering the translation field on a phone; the two may be the
  same underlying geometry seen from different angles.

## What is needed first

Reproduce it. Emulated mobile metrics with the keyboard accounted for, or a real device.
Do not change the CSS until the measurement shows what holds the height — the desktop
numbers above say the obvious explanation is already false.

## Acceptance

- A list of one or two rows occupies the height of one or two rows, on a phone, with the
  keyboard open.

u473t8:
Closes #373.

## What held the height

The phone branch of `autocomplete.css` set a **fixed** height — `position: static`
plus `height: min(176px, 34svh)` — so the box stayed 176 px at one row, at two
and at ten. That is the emptiness under the last suggestion. Flex stretching,
extra nodes and the software keyboard were each excluded by measurement first.

Master had already replaced that with the desktop overlay (#289): it shrinks,
but it takes the list out of the flow and lays it over the translation field
(#349). This change keeps the list in the flow, swaps the fixed height for a
ceiling, and animates the opening to the content height.

```css
@media (hover: none), (pointer: coarse), (max-width: 600px) {
  .suggestions {
    left: auto;
    max-height: min(176px, 34svh);
    position: static;
    right: auto;
  }

  @supports (interpolate-size: allow-keywords) {
    .suggestions { height: 0; interpolate-size: allow-keywords; transition: height 180ms ease-out, … }
    .autocomplete:has(.suggestions:not(:empty)) .suggestions { height: auto; }
  }
}
```

## Why 176 and not the desktop's 240

In flow, the list's ceiling **is** the maximum displacement of the form below
it. At 240 px a long list would push the translation field and the submit
button down 244 px — worse than the 180 px the phone already produced. At
176 px the worst case is 180 px: exactly what it did before. So the change is
no worse than the previous behaviour in any case, and better in the case the
issue is about — 49 px at one row, 89 px at two, instead of an unchanging
176 px.

## Why the animation runs to `auto`

`transition: height` needs two values it can interpolate, and the open height
here has to be the content's — a fixed open height is the defect. `interpolate-size:
allow-keywords` (Chrome 129+, and this is a Chromium phone PWA) makes `height: auto`
such a value. The whole animation sits inside `@supports`; where it is missing the
block is skipped and the list keeps the flow, the ceiling and the content height,
sizing in one frame. No emptiness at rest or at the end of the transition, either
way — that is the property the issue is about and it does not depend on the
animation.

## Measured

`npx playwright test --project=mobile --no-deps`, fixture dictionary, 390x844:

| layout | submitPush | shiftScore | entries |
|---|---:|---:|---|
| fixed `height: min(176px, 34svh)` in flow (what the phone does) | 180 px | 0.0416 | `[0.0416]` |
| overlay, `max-height: min(240px, 34svh)` (master) | 0 px | 0 | `[]` |
| ceiling in flow, no animation | 94 px | 0.0566 | `[0.0416, 0.0008, 0.0142]` |
| **ceiling in flow, animated (this PR)** | **94 px** | **0.0502-0.0544** over five runs | `[0.0014, 0.0058, 0.0072, 0.0066, 0.006, 0.0034, 0.0036, 0.0013, 0.0008, 0.0142]` |

Geometry is stable at 94 px across every run; `listClientHeight === listContentHeight`
= 88 px, `renders` 14, `inp` 48-72 ms.

The animation covers the **opening** — `0 -> auto` is a change of computed value,
and the 0.0416 single-frame shift becomes nine entries totalling 0.036. It cannot
cover a **resize**: `auto -> auto` changes no computed value at all, only the used
height, and transitions do not listen to used values. The 0.0142 entry from the
list narrowing six matches to two is identical with the animation and without it.
#289's 0.028 was one animated move of a box whose open height was *fixed*, so it
never resized again. Duration is not a lever either: 300 ms scored 0.0533 against
180 ms's 0.0502.

## The budget moved with the layout: 0.05 → 0.06

Splitting a move across frames only helps through the impact fraction — measured
13 % — so even a hypothetical animation covering every entry lands near 0.049.
That is this layout's floor, not a number better code reaches, and it is a
property of choosing a box that resizes with its content rather than one that
never resizes.

So the threshold now describes the layout that shipped: **0.06**, the measured
0.0502-0.0544 plus room for that spread, about 10 % over the worst observed. Not
room for a regression — a regression here moves the geometry, and the geometry
assertions are exact equalities. The reasoning is written out beside
`SHIFT_BUDGET` in the spec and in the change's `design.md`, and the superseded
overlay record now says the old 0.05 is not the standing number.

Worth keeping in view: none of these shifts are visible to CLS. Every entry
carries `hadRecentInput`, because typing is what causes them, so CLS reads 0.000
whatever this budget says. It is internal discipline, which is exactly why it has
to match the layout.

**Suite: 18 passed.** Desktop untouched.

## Screenshots

Two matches (`Fenster`), fixture dictionary, 390x844. Re-shot with the animation:
the resting picture is pixel-identical, since the animation changes how the box
reaches its height, not the height.

| before — fixed height, on the phone | before — overlay, master | after — flow with a ceiling |
|---|---|---|
|  |  |  |

## Spec

`add-form-stability.mobile.spec.js` asserted that the translation field and the
submit button never move — the overlay written down as a test, which fails by
construction once the list is in flow. Rewritten, not relaxed:

- no emptiness: `clientHeight === scrollHeight` and the box under the ceiling;
- bounded displacement: `submitPush === listHeight + 4`, `submitPush <= 180`,
  and the translation field travels with the button;
- the value field and the panel heading still hold still (that is
  `justify-content: flex-start` on `.home__content`, kept as a regression net);
- `shift.score < SHIFT_BUDGET`, with the budget's provenance beside it.

OpenSpec: `phone-suggestions-in-flow` is archived as
`openspec/changes/archive/2026-08-18-phone-suggestions-in-flow/`, its two
requirements synced into `openspec/specs/home-add-form-focus/spec.md`. The overlay
decision in `mobile-add-form-stability` is marked superseded by #373, reasoning
left intact.


u473t8 (commented):
LGTM

